PR #33, opened 2021-06-22
London Class 7- Oksana Kostenko- Full-Stack-Project-Assessment

oksygenn:


client/src/components/AddVideoForm.js
```diff
@@ -0,0 +1,32 @@
...
+      <label htmlFor="movie-title">
+        Title
+        <input ref={titleRef} type="text" name="movie-title" id="movie-title" />
+      </label>
+      <label htmlFor="movie-url">
+        URL
+        <input ref={urlRef} type="url" name="movie-url" id="movie-url" />
+      </label>
+      <br />
+      <button>CANCEL</button>
```
ARTurnerGit, line 18: How come this button doesn't have any click handler?

client/src/components/SingleVideo.css
```diff
@@ -0,0 +1,55 @@
...
+  text-align: center;
+  box-shadow: 1px 4px 18px -1px rgba(0, 0, 0, 0.4);
+  padding: 1rem;
+}
+
+.votes-container {
+  display: flex;
+  justify-content: space-around;
+  align-items: center;
+  /* background-color: aquamarine; */
```
ARTurnerGit, line 11: It's good to get into the habit of removing commented out code that you're not using - if you commit your work regularly you can _always_ go back to what you had before, so don't be scared to take stuff out.

client/src/components/SingleVideo.js
```diff
@@ -0,0 +1,61 @@
...
+
+  return (
+    <div className="video-card">
+      <h3>{title}</h3>
+      <div className="votes-container">
+        <div>
+          <FaThumbsUp
+            onClick={handleUpVote}
+            className={votes === 1 ? "voted thumb-icon" : "thumb-icon"}
+            size={30}
```
ARTurnerGit, line 25: It's fine setting the sizes individually like this, but as the app grows this would be a good candidate for becoming a variable eg `const iconSize = 30` that you can set in one place (towards the top of the `SingleVideo` function) and then use that `iconSize` variable to set both this and the thumbs down icon size. 

Then when you change your mind on icon sizes, it's one thing to change in one place.

client/src/components/SingleVideo.js
```diff
@@ -0,0 +1,61 @@
+import { useState } from "react";
+import { FaThumbsUp, FaThumbsDown } from "react-icons/fa";
+import "./SingleVideo.css";
+
+const SingleVideo = ({ video, handleDeleteVideoClick }) => {
+  const { title, url, rating, id } = video;
```
ARTurnerGit, line 6: Good use of destructuring

client/src/components/Videos.js
```diff
@@ -0,0 +1,83 @@
+import { useState } from "react";
+import staticData from "../exampleresponse.json";
+import AddVideoForm from "./AddVideoForm";
+import Search from "./Search";
+import SingleVideo from "./SingleVideo";
+import "./Videos.css";
+
+const Videos = () => {
+  const sortedVideosData = staticData.sort((a, b) => b.rating - a.rating);
```
ARTurnerGit, line 9: When this function `Videos` is rerendered, it will run the entire function top to bottom each time.

This means that line 9, the sorting, is _also_ going to be run each time. It seems to me you're only interested in sorting this once though, then storing it in the `videosData` state. 

Two choices for how to achieve this. 

1. Easiest way - just move line 9 up so that it comes before the `Videos` function
2. Neater way - you can initialise useState using a function like this:
```const [variable, setVariable] = useState(() => staticData.sort(...))```

This is called Lazy initial state and is explained here: [https://reactjs.org/docs/hooks-reference.html#usestate](https://reactjs.org/docs/hooks-reference.html#usestate)

File: client/src/components/Videos.js [hunk 6]
```diff
@@ -0,0 +1,83 @@
...
+  const [videosData, setVideosData] = useState(sortedVideosData);
+  const [searchValue, setSearchValue] = useState("");
+  const [showError, setShowError] = useState(false);
+
+  // delete video (for SingleVideo)
+  const handleDeleteVideoClick = (id) => {
+    const newVideosData = videosData.filter((video) => {
```
ARTurnerGit, line 16: Yes! A filter!

Don't forget that for arrow functions like this you can use an implicit return by writing:
`.filter(video => video.id !== id)`
It's called an implicit return because it removes the `return` and the curly braces

client/src/components/Videos.js
```diff
@@ -0,0 +1,83 @@
...
+      const newVideoObj = {
+        id: Date.now(),
+        title,
+        url,
+        rating: 0,
+      };
+
+      const newVideosData = [...videosData]; // creates copy of videosData
+      newVideosData.push(newVideoObj);
+      setVideosData(newVideosData);
```
ARTurnerGit, line 47: It's good that you're using spread here. Perhaps a neater way of doing it would be to say:
`setVideosData([...videosData, newVideoObj])`
It's equivalent to what you have here, just a little more concise.

client/src/components/AddVideoForm.js
```diff
@@ -0,0 +1,32 @@
+import { useRef } from "react";
+
+const AddVideoForm = ({ handleAddVideo }) => {
+  const titleRef = useRef(null);
+  const urlRef = useRef(null);
```
ARTurnerGit, line 5: This is an interesting way of doing it. It's completely valid. Another approach would be to setup a piece of state that is constantly updated to reflect what the user has input (so making your input forms "controlled" inputs).

client/src/components/Videos.js
```diff
@@ -0,0 +1,83 @@
...
+        </p>
+      )}
+      <AddVideoForm handleAddVideo={handleAddVideo} />
+      <Search handleSearch={handleSearch} searchValue={searchValue} />
+      <section className="videos">
+        {filteredVideos.map((video, index) => {
+          return (
+            <SingleVideo
+              video={video}
+              key={index}
```
ARTurnerGit, line 72: Can you think of anything better in the data that you could use as the key here?

ARTurnerGit (commented):
Looks really good Oksana. All my comments are really minor things (or discussion points to get you thinking).

Well done.

client/src/components/AddVideoForm.js
oksygenn: I'm still not sure if I want to have that button, that's why it doesn't have a click handler :)

client/src/components/SingleVideo.js
oksygenn: Thanks :)

client/src/components/AddVideoForm.js
oksygenn: I'll remove it for now then 👍

PR #47, opened 2021-11-26
Full Stack Assessment Project

dharmaguadeloupe:


server/server.js
```diff
@@ -1,15 +1,58 @@
...
+});
+
+//Get video by ID
+app.get("/:videoId", (req, res) => {
+  const videoId = req.params.videoId;
+  const videoWithId = videos.find((video) => video.id === videoId);
+  videoWithId
+    ? res.send(videoWithId)
+    : res.status(404).send({ msg: "Video doesn't exist" });
+});
```
KarenPudner, line 25: Goo use of error message and status code.

server/server.js
```diff
@@ -1,15 +1,58 @@
...
+
+//Add New Video
+let tempId = Math.floor(Math.random() * 10000);
+app.post("/", (req, res) => {
+  const title = req.body.title;
+  const url = req.body.url;
+
+  if (!title || !url || !url.includes("youtube") || url.includes("watch?v=")) {
```
KarenPudner, line 33: Should be !url.includes("watch?v=").

server/server.js
```diff
@@ -1,15 +1,58 @@
...
+    return res
+      .status(400)
+      .send({ result: failure, msg: "Video could not be saved" });
```
KarenPudner, line 36: You need to put failure in quotes (as you have done elsewhere)

server/server.js
```diff
@@ -1,15 +1,58 @@
...
+    title: title,
+    url: url,
+  };
+  videos.push(newVideoToAdd);
+  res.send({ id: newVideoToAdd.id });
+});
+
+app.delete("/:videoId", (req, response) => {
+  const videoId = req.params.videoId;
+  const videoIndex = videos.findIndex((video) => video.id === videoId);
```
KarenPudner, line 49: I also had to change "===" to "==" here to make this work.

client/src/App.css
```diff
@@ -1,3 +1,51 @@
...
+.addVideoContainer {
+  text-align: center;
+}
+
+.card-group {
+  display: flex;
+  flex-flow: row wrap;
+  justify-content: space-around;
+}
+.card {
```
Nomes27, line 23: your iframe is overhanging the card at the mo, if you set the width of the iframe to 100% should fit nicer :)

client/src/components/AddVideo.js
```diff
@@ -0,0 +1,70 @@
...
+    const newVideoData = {
+      title: title,
+      url: url,
+    };
+    setVideoData(videoData.concat(newVideoData));
+  };
+
+  return (
+    <section className="addVideoContainer">
+      <h4 onClick={() => setAddVideo(!addVideo)}>Add Video</h4>
```
Nomes27, line 20: a button would be more appropriate here

client/src/components/Buttons.js
```diff
@@ -0,0 +1,37 @@
+import React, { useState } from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faThumbsUp, faThumbsDown } from "@fortawesome/free-solid-svg-icons";
+
+const Buttons = () => {
```
Nomes27, line 5: nice, I like how you've broken these into a separate component!

client/src/components/Buttons.js
```diff
@@ -0,0 +1,37 @@
...
+  const [totalVotes, setTotalVotes] = useState(0);
+  const increaseVote = () => {
+    setTotalVotes(totalVotes + 1);
+  };
+  const decreaseVote = () => {
+    if (totalVotes >= 0) {
+      setTotalVotes(totalVotes - 1);
```
Nomes27, line 12: at the moment the decrease is allowing votes to go to -1, rather than stopping at 0 (using > rather than >= might be a good idea)

client/src/components/Buttons.js
```diff
@@ -0,0 +1,37 @@
...
+          onClick={increaseVote}
+          style={{ color: "green" }}
+        />
+        <p className="votes">{totalVotes} votes</p>
+        <FontAwesomeIcon
+          icon={faThumbsDown}
+          size="3x"
+          onClick={decreaseVote}
+          style={{ color: "red" }}
+        />
```
Nomes27, line 31: nice use of font awesome icons! I would wrap each in a button, since clicking them has some functionality (and give the button an aria-label for accessibility)

client/src/components/SearchBar.js
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+
+const SearchBar = ({ videoData, setVideoData }) => {
+  const handleSearch = (e) => {
+    const searchVideoData = videoData.filter((video) =>
+      video.title.toLowerCase().includes(e.target.value.toLowerCase())
+    );
+    setVideoData(searchVideoData);
+  };
```
Nomes27, line 9: this works really well, I like how it filters as you type and removes the need to click a button! To improve it might be nice to set the video data back to the full list when the search input returns to an empty string

client/src/components/VideoCards.js
```diff
@@ -0,0 +1,42 @@
+import React from "react";
+import Buttons from "./Buttons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faTrashAlt } from "@fortawesome/free-solid-svg-icons";
+
+const VideoCards = ({ videoData, setVideoData }) => {
```
Nomes27, line 6: nice that you've separated these into a component, you could go further & have a component for each video card

client/src/components/VideoCards.js
```diff
@@ -0,0 +1,42 @@
...
+  const handleDelete = (id) => {
+    const filteredVideos = videoData.filter((video) => video.id !== id);
+    setVideoData(filteredVideos);
+  };
```
Nomes27, line 10: this delete function works really well :)

client/src/App.js
```diff
@@ -1,20 +1,54 @@
...
+          <VideoCard
+            title={video.title}
+            url={video.url}
+            rating={video.rating}
+            id={video.id}
+            key={index}
+            deleteCard={handleDelete}
+          />
+        ))}
+      </div>
```
Nomes27, line 51: nice, this works much better having a separate VideoCard component!

client/src/components/AddVideo.js
```diff
@@ -9,41 +8,39 @@ const AddVideo = ({ videoData, setVideoData }) => {
...
-              Title
-              <input
-                className="input"
-                name="title"
-                type="text"
-                required
-                placeholder="title"
-                onChange={(e) => setTitle(e.target.value)}
-              ></input>
-            </label>
```
Nomes27: I would always try to have a label for inputs if possible (helps massively with accessibility) if not, you could add an aria-label to the input for example `aria-label="title"`

client/src/components/Buttons.js
```diff
@@ -8,27 +8,35 @@ const Buttons = () => {
...
+        <button>
+          <FontAwesomeIcon
+            aria-label="button"
+            role="button"
+            icon={faThumbsUp}
+            size="3x"
+            onClick={increaseVote}
+            style={{ color: "green" }}
+          />
+        </button>
```
Nomes27, line 28: this is good, much better as a button! With your aria-label here if you instead add it to the button element and give it a name which conveys meaning to screen readers (as there is only an icon and no descriptive text), for example here `aria-label="upvote"` would work

client/src/components/SearchBar.js
```diff
@@ -2,14 +2,15 @@ import React from "react";
...
-    const searchVideoData = videoData.filter((video) =>
-      video.title.toLowerCase().includes(e.target.value.toLowerCase())
-    );
-    setVideoData(searchVideoData);
+    if (e.target.value.length > 0) {
+      const searchVideoData = videoData.filter((video) =>
+        video.title.toLowerCase().includes(e.target.value.toLowerCase())
+      );
+      setVideoData(searchVideoData);
+    }
```
Nomes27, line 10: this is close to resetting the videos when the input string is empty - you might want to look at adding in an else condition, so that if a user backspaces to return to an empty string, it shows all videos (for example if you imported your original video data in this component, you could setVideoData back to this original data if e.target.value.length became 0.

server/server.js
```diff
@@ -1,15 +1,92 @@
...
+    .query("SELECT * FROM videos WHERE id=$1", [videoId])
+    .then((db_res) => {
+      if (db_res.rowCount > 0) res.status(200).json(db_res.rows);
+      else
+        res.status(404).json({
+          result: "failure",
+          message: "Video could not be found",
+        });
+    })
+    .catch((db_err) => res.status(500).send("server error"));
```
KarenPudner: Good use of appropriate error messages

server/server.js
```diff
@@ -1,15 +1,92 @@
...
+    if (db_res.rows.length > 0) {
+      pool.query("DELETE FROM videos WHERE id=$1", [id], (db_err, db_res) => {
+        if (db_err) {
+          res.status(500).send(JSON.stringify(db_err.message));
+        } else {
+          res.json(db_res.rows);
+        }
+      });
+    } else {
+      res.status(400).send("Unable to locate the video with the id provided");
```
KarenPudner: Again, good work in identifying client side error and server side error and having different response status codes.

KarenPudner (approved):
Good work, and well done on getting the database functionality working.

PR #114, opened 2022-06-15
London Class8/Laleh Koupaei/Full-Stack-Project-Week2

lalehkoupa:
Level 250 completed

-----
[View rendered server/node_modules/fsevents/README.md](https://github.com/lalehkoupa/Full-Stack-Project-Assessment/blob/week2/server/node_modules/fsevents/README.md)


client/src/App.js
```diff
@@ -1,11 +1,11 @@
...
+import React from "react";
+import HomePage from "./component/HomePage";
 
 function App() {
   return (
     <div className="App">
-      <header className="App-header">
-        <h1>Video Recommendation</h1>
-      </header>
+      <HomePage />
```
Awyssa, line 8: Very nice! Later on, will we start to use [react-router-dom](https://v5.reactrouter.com/web/guides/quick-start) and keeping your App.js file as clean and simple as possible is very good practice. Nice one Laleh!!!

client/src/component/Addvideo.jsx
```diff
@@ -0,0 +1,77 @@
+import React, { useState } from "react";
+
+function AddVideo({ videoData }) {
```
Awyssa, line 3: In React, we always want to use arrow functions, not function declaration. I would use `const addVideo = () => {}` instead 😊

client/src/component/Addvideo.jsx
```diff
@@ -0,0 +1,77 @@
...
+  const [allData, setAllData] = useState(videoData);
+
+  const handleShowForm = () => {
+    SetShowForm(!showForm);
+  };
+  const handleAddVideo = (event) => {
+    event.preventDefault();
+
+    const newVideoData = {
+      id: Math.floor(Math.random() * 1000),
```
Awyssa, line 16: This is a nice idea, but we could still run into an issue in which we create multiple video objects with the same id.

Another solution would be to get the last index of the videoData array, and increment the id of this index. so for example `id: videoData[videoData.length -1].id + 1`. Or even better, as this looks quite confusing, we can use lowdash and do `_.last(videoData).id + 1`.

Awyssa: Let me know if that doesn't make sense and we can discuss it in our next video call :)

client/src/component/Addvideo.jsx
```diff
@@ -0,0 +1,77 @@
...
+      title,
+      url,
+    };
+
+    fetch(videoData, {
```
Awyssa, line 21: We are doing a `fetch` here, but we are doing nothing with the response. The code is also not asynchronous by using async/await or Promise chaining (.then().then().catch().finally())

If the fetch fails or is successful, we have no way of knowing. I would suggest here we use async/await and save the response to a const. With that, we can let the user know whether they added the video successfully, or if it failed.

We should also try to introduce try/catch here too.

So for example
```
try
{
   const res = fetch(....

    if(res.status !== 200)
        throw new Error("Video upload failed!")

     setAllData({ ...allData, ...form})
     setForm({
        id: "",
        title: "",
        ...
{
catch(error)
{
  console.log("error: ", error)
}
```

I've not written this out completely, but just as guidance on how I would do this. Again, should this not make sense we can discuss over a video call again soon :)

client/src/component/Video.jsx
```diff
@@ -0,0 +1,49 @@
+import React, { useState, useEffect } from "react";
+
+import DeleteButton from "./DeleteButton";
+import Rating from "./Rating";
+
+function Video({ videoData }) {
```
Awyssa, line 6: In React we use arrow functions, not function declaration :)

client/src/component/Video.jsx
```diff
@@ -0,0 +1,49 @@
...
+  const [removedId, setRemovedId] = useState([]);
+
+  function handleClick(id) {
```
Awyssa, line 9: In React we use arrow functions, not function declaration :)

Sorry Laleh, I know I am being annoying 😅  I just want to keep saying it so we remember :)

client/src/component/Rating.jsx
```diff
@@ -0,0 +1,24 @@
+import React, { useState } from "react";
+
+function Rating(props) {
```
Awyssa, line 3: Being annoying again, but again, use arrow functions 😅

client/src/component/Rating.jsx
```diff
@@ -0,0 +1,24 @@
...
+  const [vote, setVote] = useState(props.rating);
+
+  function handleClickUpVote() {
```
Awyssa, line 6: I would make one function and pass an argument in the function on whether to add down or to add up. For example. 

```
const handleVote = (upVote) => 
{
  setVote( upVote ? vote + 1 : vote - 1)
}
...
<button onClick={() => handleVote(true)}>up vote</button>
<button onClick={() => handleVote(false)}>down vote</button>

File: client/src/component/HomePage.jsx [hunk 9]
```diff
@@ -0,0 +1,28 @@
+import React, { useState, useEffect } from "react";
+import Video from "./component/Video";
+import Header from "./component/Header";
+import AddVideo from "./component/AddVideo";
+
+function HomePage() {
+  const [videos, setVideos] = useState([]);
+
+  const getData = async () => {
```
Awyssa, line 9: Woo!!! Async/await and an arrow function!!! 🥳 
Love it Laleh!!! Great work!!!

client/src/component/HomePage.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+  useEffect(() => {
+    getData();
+  }, []);
+
+  return (
+    <div>
+      <Header />
+      <AddVideo videoData={videos} />
+      <Video videoData={videos} />
```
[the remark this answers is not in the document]
Awyssa: If I was to be super pedantic, you could replace the <div> with <>. This will lower the bundle size and DOM elements, making our website lighter and faster :)

client/src/component/Header.jsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+function Header() {
```
Awyssa, line 3: Being very annoying again, make this an arrow function :p

Awyssa:
Overall, this is some amazing work @lalehkoupa!!! 

I know changing between Node, Express, and the backend, back to React, HTML, CSS, and the frontend can be super confusing, and we feel like we forget a lot of what we learned. 

But you are doing such a good job and this is some really really good code!!! I am being very annoying and picky, just because I feel very comfortable pushing you very hard as you are doing so well! So when you see my comments just know it's not because you are doing anything wrong, you are doing great! I just want you to get even better!

As always, you should be very proud of this amazing work, and I hope my comments help 🥳

lalehkoupa:
@Awyssa Thank you so much Michael,I tried to apply all of the changes into my code, many thanks for reviewing my code and your helpful comment

PR #116, opened 2022-06-17
London Class 8 - Hamra Khan - Full Stack Project - Week 1 - Front End

HamraKhan:
null

-----
[View rendered server/node_modules/@sindresorhus/is/readme.md](https://github.com/HamraKhan/Full-Stack-Project-Assessment/blob/level-100/server/node_modules/@sindresorhus/is/readme.md)


client/src/AddVideo.js
```diff
@@ -0,0 +1,82 @@
...
+            <button
+              type="button"
+              className=" btn add-video-buttons add-btn"
+              onClick={addVideoObj}
+            >
+              Add
+            </button>
+          </div>
+
+          <div className="cancel-button">
```
ewintram, line 68: Well done for adding a cancel button.

client/src/AddVideo.js
```diff
@@ -0,0 +1,82 @@
...
+
+    function addVideoObj() {
+      const videoObj = {title: videoTitleInput, url: videoUrlInput};
+      props.AddVideo(videoObj);
+      setVideoTitleInput("");
+      setVideoUrlInput("");
+    }
+
+    function handleReset() {
+      Array.from(document.querySelectorAll("input")).forEach(
```
ewintram, line 23: which other inputs does this handle? Can you just use the state hooks to reset `title` & `url` inputs?

client/src/AddVideo.js
```diff
@@ -0,0 +1,82 @@
...
+      setVideoTitleInput(event.target.value);
+    }
+
+    function handleUrlChange(event) {
+      setVideoUrlInput(event.target.value);
+    }
+
+    function addVideoObj() {
+      const videoObj = {title: videoTitleInput, url: videoUrlInput};
+      props.AddVideo(videoObj);
```
ewintram, line 17: I think `addVideo` here is a function, not a component, so it should be a lowercase `a` at the beginning

client/src/VideoCards.js
```diff
@@ -0,0 +1,52 @@
...
+
+  /* remove any video by id */
+  function removeVideo(id) {
+    const updatedVideoList = videoList.filter((video) => video.id !== id);
+    setVideoList(updatedVideoList);
+  }
+
+  /* add any video to list */
+  function addVideo(videoObj) {
+    setVideoList([videoObj, ...videoList]);
```
ewintram, line 18: nice use of the spread operator to copy the array!

client/src/VideoCards.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return (
+    <div className="video-cards">
+      <AddVideo AddVideo={addVideo} />
```
ewintram, line 23: the prop `AddVideo` should be in camel-case, so with a lowercase `a`. Only component names being with a capital letter
```
<AddVideo addVideo={addVideo} />
```

client/src/VideoCards.js
```diff
@@ -0,0 +1,52 @@
...
+              {/* video title */}
+              <div className="card-title">
+                <h2>{videoObj.title}</h2>
+              </div>
+
+              <Video
+                /* embedded video */
+                videoObj={videoObj}
+                /* delete button */
+                removeVideo={() => {
```
ewintram, line 38: nice inline callback function

client/src/VideoCards.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return (
+    <div className="video-cards">
+      <AddVideo AddVideo={addVideo} />
+      <div className="cards">
+        {videoList.map((videoObj, index) => {
+          return (
+            /* video contents */
+            <div className="card-container" key={index}>
```
ewintram, line 28: well done remembering to add the key!

client/src/Videos.js
```diff
@@ -0,0 +1,29 @@
+import React from "react";
+
+function Video(props) {
```
ewintram, line 3: If you prefer, you can destructure the props as parameters to the component, to avoid having to write `props.videoObj` within the component. i.e.
```
function Video({ videoObj, removeVideo }) {
```

client/src/Videos.js
```diff
@@ -0,0 +1,29 @@
...
+    const videoUrl = props.videoObj.url.replace("watch?v=", "embed/");
+
+    return (
+      <div className="videos">
+        {/* delete button */}
```
ewintram, line 8: Just a tip - in general I would recommend avoiding leaving comments in your code. Your code should describe itself based on well named components, functions and classes. As your code changes, your comments may become out of date (nothing breaks if you leave an old, invalid comment in there), so commenting can become unreliable. If you feel like you need to leave a comment because you can't remember what a piece of code is doing, chances are it could be extracted into a named component or function to be more descriptive.

ewintram: In this case though, I don't think you need to change the code, it speaks for itself

client/src/VoteButtons.js
```diff
@@ -0,0 +1,46 @@
+import React, { useState } from "react";
+
+function VoteButtons() {
+    const [vote, setVote] = useState(0);
+
+    function incrementCount() {
+        setVote(vote + 1);
+    }
+
+    function decrementCount() {
```
ewintram, line 10: great use of hooks

ewintram (commented):
the site looks super & all the frontend functionality works great

PR #134, opened 2022-06-26
London Class 8 - Hamra Khan - Full Stack Project - Week 3 - Database

HamraKhan:


server/videoModel.js
```diff
@@ -0,0 +1,12 @@
+const mongoose = require("mongoose");
+
+const schema = mongoose.Schema({
+  id: Number,
+  title: String,
+  url: String,
+  rating: Number
+});
+
+const Videos = mongoose.model("videos", schema, "videos");
```
ewintram, line 10: it is convention that model names are singular and table names are plural, i.e. 
```
const Video = mongoose.model("video", schema)
```
according to the docs here https://mongoosejs.com/docs/models.html

server/videoService.js
```diff
@@ -0,0 +1,43 @@
+const { json } = require("express");
+const { Videos } = require("./videoModel");
+
+function getVideos(id) {
+    if(id) { return Videos.find({id})}
+
+    return Videos.find();
```
ewintram, line 7: nice handling of both cases with and without an id

server/videoService.js
```diff
@@ -0,0 +1,43 @@
...
+        } else {
+            return {
+                result: "failure",
+                message: "Video could not be deleted because it does not exist",
+            };  
+        }
+    });
+}
+
+module.exports = { getVideos, addVideo, deleteVideo };
```
ewintram, line 43: great job

PR #143, opened 2022-12-08
ZA2 - Anthony Mogotlane - Full-Stack-Project-Assessment - W1, W2, W3

AnthonyMogotlane:
Up and down vote buttons are working, a video can be added and a video can be remove from a list of videos


SallyMcGrath (changes requested):
👋 Hi @AnthonyMogotlane 

There are 180 files in this PR. Which ones would you like mentors to review? Would it be a good idea to commit only the pertinent files in your change list? 😉 

PR #144, opened 2022-12-09
ZA2-Chezlin Benson-Full-Stack-Project-Assessment

chezlinbenson:


SallyMcGrath (changes requested):
👋 Hi @chezlinbenson 

There are 165 files in this PR. Which ones would you like mentors to review? Would it be a good idea to commit only the pertinent files in your change list? 😉 

PR #145, opened 2022-12-09
ZA2-Nishka Kisten-Full-Stack-Project-Assessment

Nishka-Kisten:


SallyMcGrath (changes requested):
👋 Hi @Nishka-Kisten 

There are over 1,000 files in this PR. Which ones would you like mentors to review? Would it be a good idea to commit only the pertinent files in your change list? 😉 

PR #146, opened 2022-12-12
ZA2 - Chandre De Wet - Full Stack Assessment - level 300

chandredewet:


SallyMcGrath (changes requested):
👋 Hi @chandredewet 

There are over 5,000 files in this PR. Which ones would you like mentors to review? Would it be a good idea to commit only the pertinent files in your change list? 😉 

PR #148, opened 2022-12-17
ZA2 - Shafiek Davids - Full-Stack-Project-Assessment

shafiekdavids:
Completed  'like' and 'dislike' section.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,8 +1,14390 @@
 {
   "name": "video-recommendation",
   "version": "0.1.0",
-  "lockfileVersion": 1,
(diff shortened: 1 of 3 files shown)


SallyMcGrath (changes requested):
👋 Hi @shafiekdavids 

I can't see your work here. We had a problem recently: another trainee accidentally committed their work to the repo a while back and some trainees ended up with a solution in their fork. Everyone removed the code and added their own work instead. Did you miss the message to do this? (It's ok if you did - it's totally fixable.)

Can you upload your work please? Thanks! 😄

PR #151, opened 2023-01-08
ZA2-Christian-Mfuke-Full-Stack-Project-Assessment

mfukechristian:
Level 100 completed


SallyMcGrath (changes requested):
👋 Hi @mfukechristian 

There are over 1,000 files in this PR. Which ones would you like mentors to review? Would it be a good idea to commit only the pertinent files in your change list? 😉 

PR #161, opened 2023-01-20
WM4 - AzinYad - SQL - Week1 - Full-Stack-Project-Assessment Public

AzinYad:


client/src/App.js
```diff
@@ -1,10 +1,67 @@
...
+import VideoCard from "./component/VideoCard";
+import AddVideo from "./component/AddVideo";
+import React from "react";
+import { useState, useEffect } from "react";
+import Data from "./exampleresponse.json";
 
 function App() {
+  const [videosList, setVideosList] = useState(Data);
+  const [deletedVideoId, setDeletedVideoId] = useState(null);
+  // const [newList, setNewList] = useState({});
```
SallyMcGrath, line 11:
```suggestion

```

Make sure to remove dead code from your commits.

client/src/App.js
```diff
@@ -1,10 +1,67 @@
...
+  const [formData, setFormData] = useState({ title: "", url: "" });
+
+
+  useEffect(function removeSelectedVideo() {
+    // let targetedVideoIndex = videosList.findIndex((video) =>video.id == deletedVideoId);
```
SallyMcGrath, line 16:
```suggestion

```
Clean up before committing

client/src/App.js
```diff
@@ -1,10 +1,67 @@
...
+
+  function handleSubmit(event) {
+    event.preventDefault();
+    if (formData.title !== "" && formData.url.includes("www.youtube.com/watch?v=")){
+      let newEntryId;
+      let randomNum;
+      const generateRandomNum=()=>randomNum = Math.floor(100000 + Math.random() * 900000);
+      generateRandomNum();
+      (videosList.includes(randomNum))?generateRandomNum():newEntryId=randomNum;
+      const newRate = Math.floor(Math.random() * 9000);
```
SallyMcGrath, line 38: What is the purpose of generating a random number here in newRate?

client/src/component/VideoCard.jsx
```diff
@@ -0,0 +1,25 @@
...
+import EmbeddingVideo from "./EmbeddingVideo";
+import Vote from "./Vote";
+
+function VideoCard(props){
+   
+
+    return(
+        <section className="cards-grid">
+            { 
+            props.videosList.sort((a,b) => b.rating - a.rating).map(film=>{
```
SallyMcGrath, line 11: I like that you're thinking about sorting up here. Does the sort array method mutate the original array? What are the side effects of doing this?

https://doesitmutate.xyz/sort/

client/src/App.js
```diff
@@ -1,10 +1,67 @@
...
 function App() {
+  const [videosList, setVideosList] = useState(Data);
+  const [deletedVideoId, setDeletedVideoId] = useState(null);
+  // const [newList, setNewList] = useState({});
+  const [formData, setFormData] = useState({ title: "", url: "" });
+
+
+  useEffect(function removeSelectedVideo() {
+    // let targetedVideoIndex = videosList.findIndex((video) =>video.id == deletedVideoId);
+    let filteredVideos = videosList.filter((video) => video.id != deletedVideoId);
```
SallyMcGrath, line 17: What does your linter say about this line when you run your app locally?

How can linters help you improve your code?

SallyMcGrath (changes requested):
Really great start @AzinYad ! Thanks for this.  Lots done here! I've left a few questions and things to think about. Take a look.

I can't see your Lighthouse report. What do the requirements ask for in the readme?

(I ran Lighthouse on your code and it scored 100/100 ⭐ )

Make sure to include your report.

AzinYad:
> Really great start @AzinYad ! Thanks for this. Lots done here! I've left a few questions and things to think about. Take a look.
> 
> I can't see your Lighthouse report. What do the requirements ask for in the readme?
> 
> (I ran Lighthouse on your code and it scored 100/100 star )
> 
> Make sure to include your report.


Thanks a lot Sally ,for taking time to review my code, it means a lot to me ,
you are right I've forgot to run the lighthouse report but I will include it.

client/src/App.js
AzinYad: I just wanted to have a hypothetical rate for new videos , to sort in a descending order ,just for the website to look more real

client/src/component/VideoCard.jsx
AzinYad: Yes it mutates the original array , but I wasn't aware of the side effects as I thought we just need the descending order.and keeping the original order might not be necessary.I will look into it. thanks

client/src/App.js
AzinYad: Thanks for pointing out this Issue Sally , It was my question too
I wanted to use !==  instead of != but when I use double equal sign the code doesn't work , I wanted to ask the reason because I can't figure the reason  out

PR #164, opened 2023-01-20
WM4 - Joanna_Jurczuk - Full-Stack-Project-Assessment

joannajurczuk94:


client/src/AddNewVideo.js
```diff
@@ -0,0 +1,68 @@
...
+
+const AddNewVideo = (props) => {
+  const [title, setTitle] = useState("");
+  const [url, setUrl] = useState("");
+
+  const handleOnSubmit = (event) => {
+    event.preventDefault();
+    //Increasing of each video ID
+    const newId = (id) => {
+      return id++;
```
alicagatay, line 11: I am not sure what are you trying to do here. If you are trying to get a new id number that is 1 bigger than the old one, you can just say `return id + 1;`. This way, it will be a more clear way.

client/src/App.css
```diff
@@ -1,3 +1,42 @@
...
+  margin-bottom: 1rem;
+}
+.containerVideo {
+  background-color: red;
+  padding: 0rem 0rem 0rem 0rem;
+}
+/* .form-container {
+  margin: 0rem 6rem 3rem 6rem;
+
+} */
```
alicagatay, line 42: If you are not going to use this code part, my advise would be to just delete it. Do not get me wrong, there is nothing wrong with it, it is just cluttering the codebase.

client/src/App.js
```diff
@@ -1,11 +1,56 @@
...
+  const [addVideo, setAddVideo] = useState(false);
+
+  //Removing 1 video
+  // const deleteViedeos = (deletedViedeoId) => {
+  //   setVideoData(
+  //     allVideos.filter((video) => {
+  //       return video.id !== deletedViedeoId;
+  //     })
+  //   );
+  // }
```
alicagatay, line 18: Again, if you are not going to use this commented out code, it is best to just delete it in order to simplify your codebase.

client/src/App.js
```diff
@@ -1,11 +1,56 @@
...
+
+  const deleteViedeos = (id) => {
```
alicagatay, line 20: Check your typo's here.

client/src/RenderVideo.js
```diff
@@ -0,0 +1,55 @@
...
+  <section className="video">
+    <div>
+      <div className="title">
+        <h3>{video.title}</h3>
+        <iframe
+          width="500"
+          height="300"
+          src={`http://www.youtube.com/embed/${video.url.split("v=")[1]}`}
+          title="Video Player"
+          frameBorder="1"
```
alicagatay, line 29: The `frameBorder` attribute of the `iframe` element is deprecated, so it is best not to use it in your code,

alicagatay (changes requested):
A pretty good start Joanna. A really elegant design and the functionality also works as expected in the requirements. There are some few small issues which I have commented on my comments, but other than those, you can now star into level 200.

PR #165, opened 2023-01-20
WM4 - Abdirahim Hussein - Full-Stack-Project-Assessment

12shussein:


.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
PakizeBozkurt, line 1: .gitignore file needs to created before pushing your code!

PakizeBozkurt (commented):
There are 1082 files! 😫

.gitignore
12shussein: I did not know about .gitignore files so I would delete server and client/side node modules when pushing

PR #174, opened 2023-01-22
WM4 - Rabie Al-Ashmali - Full-Stack-Project-Assessment  

AL-ASHMALI:
…


AL-ASHMALI:
> Hi there @AL-ASHMALI please update your PR title with your name, region, and coursework title. If you take a look at the PR list on this repo, you can see how everyone else is doing it.

Sorry for that, all should be done now

server/server.js
```diff
@@ -1,15 +1,60 @@
 const express = require("express");
 const app = express();
+const cors = require("cors");
 const port = process.env.PORT || 5000;
+const videos = require("./exampleresponse.json");
```
Louuu, line 5: This file seems to be missing from your server directory, which causes the error `Error: Cannot find module './exampleresponse.json'` when starting the server

AL-ASHMALI: Thank you for your help. It is working now

PR #175, opened 2023-01-23
ZA2-Tersia Koetzee-Full-Stack-Project-Assessment

tersiakoetzee:


SallyMcGrath (changes requested):
Hi @tersiakoetzee . There are 1494 files in this PR. Which ones were you hoping mentors review? 😉

Would it be a good idea to open a PR only with the pertinent files? (Never git add . in a codebase)

tersiakoetzee:
> Hi @tersiakoetzee . There are 1494 files in this PR. Which ones were you hoping mentors review? 😉
> 
> Would it be a good idea to open a PR only with the pertinent files? (Never git add . in a codebase)

Hi Sally thank you for that I will get to fixing it.

PR #180, opened 2023-01-24
Za2-Advocate-Maroga-Full-stack-project-assessment

AdvocateM:


client/src/App.js
```diff
@@ -1,11 +1,104 @@
+import React, { useState, useEffect } from "react";
 import "./App.css";
+// import SampleData from "./SampleData/exampleresponse.json";
+import ListVideos from "./components/ListVideos";
+import Nav from "./components/Nav";
+import FormDisabled from "./components/FormDisabled";
 
 function App() {
+  const [VideosInfo, setVideosInfo] = useState([]);
+  const [load, isLoaded] = useState(false);
```
DomVinyard, line 10: the convention for state variable names is to have the same name with `set` prepended for the setter, so in this case either

```
const [load, setLoad] = useState(false);
```

or

```
const [isLoaded, setIsLoaded] = useState(false);
```

client/src/App.js
```diff
@@ -1,11 +1,104 @@
...
+      .catch((error) => {
+        console.log(error);
+      });
+
+  }, [load]);
+
+  console.log(VideosInfo);
+
+  const addVideo = (title, url) => {
+    let newVideo = {
```
DomVinyard, line 48: prefer `const` to `let`

client/src/App.js
```diff
@@ -1,11 +1,104 @@
...
+  // remove a video prop function
+  const removeVideo = (id) => {
+    const videoCopy = [...VideosInfo];
+    console.log(`removed video with id:${id}`);
+    const index = videoCopy.findIndex((video) => {
+      return video.id === id;
+    });
+    videoCopy.splice(index, 1);
+    setVideosInfo(videoCopy);
+  };
```
DomVinyard, line 79: There is no problems with this approach, but it may be more readable if you simplify 

```
const removeVideo = (id) =>  {
    setVideosInfo(VideosInfo.filter(video => video.id !== id))
}
```

client/src/components/FormDisabled.js
```diff
@@ -0,0 +1,69 @@
+import { React, useState } from 'react';
+
+export default function FormDisabled(props) {
```
DomVinyard, line 3: What exactly is your `FormDisabled` component doing, it's very similar to AddVideo.js? Can you think of a way to reduce this duplication

DomVinyard (dismissed):
Nice work, solid implementation with a few areas to clear up

PR #181, opened 2023-01-24
WM4 - Full-Stack-Project-Assessment - Robert_Csonka

RobCso:


client/src/VideosList.js
```diff
@@ -0,0 +1,44 @@
...
+                <p>{item.title}</p>                
+                < VoteButton item={item}/>
+                <iframe
+                  width="560"
+                  height="315"
+                  src={`https://www.youtube.com/embed/${videoIdExtract(
+                    item.url
+                  )}`}
+                  title="YouTube video player"
+                  frameBorder=""
```
alicagatay, line 33: `frameBorder` attribute if the `iframe` component is deprecated, so it is better not to use it in your code.

client/src/VideosList.js
```diff
@@ -0,0 +1,44 @@
+import React from "react";
+import VoteButton from "./VoteButton";
+
+
+const VideosList = ({videoData}) => {
+    //console.log(videoData)
+    const videoIdExtract = (url) =>{
+        var regExp =
+          /^.*((youtu.be\/)|(v\/)|(\/u\/\w\/)|(embed\/)|(watch\?))\??v?=?([^#&?]*).*/;
```
alicagatay, line 9: Always use `let` or `const` instead of `var`.

Also, I can see that you have used a regular expression in this particular code. Can you please explain why have you used this regular expression? What is it doing and what are the benefits and disbenefits of using a regex here?

client/src/VideosList.js
```diff
@@ -0,0 +1,44 @@
...
+        var match = url.match(regExp);
```
alicagatay, line 10: Always use `let` or `const` instead of `var`.

client/src/Form.js
```diff
@@ -0,0 +1,70 @@
...
+      e.preventDefault();
+      let newVideo = {
+        id: Date.now(),
+        title: title,
+        url: url,
+        rating: rating
+      }
+      function matchYoutubeUrl(url) {
+        var p =
+          /^(?:https?:\/\/)?(?:m\.|www\.)?(?:youtu\.be\/|youtube\.com\/(?:embed\/|v\/|watch\?v=|watch\?.+&v=))((\w|-){11})(?:\S+)?$/;
```
alicagatay, line 19: Always use `let` or `const` instead of `var`.

Also, I can see that you have used a regular expression in this particular code. Can you please explain why have you used this regular expression? What is it doing and what are the benefits and disbenefits of using a regex here?

alicagatay (changes requested):
Good work Robert. You have created a simple interface and almost accomplished all of the requirements for level 100. The only feature left for this level is, you need to create a button for each video that will delete the video both from the list and from the user interface. After that, You will be ready to continue with level 2.

server/server.js
```diff
@@ -1,15 +1,99 @@
...
+  }
+  res.send(findVideo);
+});
+
+// POST
+
+const compulsoryFields = ["title", "url", "rating"];
+
+app.post("/", (req, res) => {
+  if (!compulsoryFields.every((cf) => req.body.hasOwnProperty(cf))) {
```
Know-Thyself, line 52: Although group validation would work just fine, breaking it down and validating the title, the url and the rating separately makes your server response more specific and less prone to errors.

client/src/VideosList.js
RobCso: Hi Ali
Thank you for making an effort looking at my work.
In this case I have used this regex to help extract the YouTube video ID from the url is I needed it as the iframe src value

client/src/Form.js
RobCso: In this case I have used the regex to help validate the URL that is submitted with the form as a valid YouTube link.

PR #183, opened 2023-01-24
WM4-DEWAYNE SEWELL-FULL STACK PROJECT ASSESSMENT WEEK2

dewayne51:


client/src/VideoAdd.js
```diff
@@ -0,0 +1,59 @@
...
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify(newVideo)
+    });
+    const data = await response.json();
+    setvideos(data);
+  } catch (error) {
+    console.error('Error:', error);
+  }
+}; */
```
alicagatay, line 32: If you are not using this commented out code, the best thing to do with it would be deleting it. Most of the times, when we comment out a code group like this, it always gets forgotten, which often makes unwanted clutter in your codebase.

client/src/buttons/DeleteButton.js
```diff
@@ -1,8 +1,15 @@
...
 
-function DeleteButton() {
+
+function DeleteButton({video, setvideos, videos})  {
   return (
-    <button class="p-2 mb-1 bg-danger text-white delete-button">Delete</button>
+    <button onClick={() => {
+      const filteredVideos = videos.filter((v) => {
+        if (video.id !== v.id)
+        {return (v)};
```
alicagatay, line 9: You could just simplify this code such as:

`return video.id !== v.id;`

There is no need for the if statement here.

client/src/buttons/LikeIcon.js
```diff
@@ -1,15 +1,64 @@
...
+        fill="currentColor"
+        class="bi bi-hand-thumbs-down-fill"
+        viewBox="0 0 16 16"
+      >
+        <path d="M6.956 14.534c.065.936.952 1.659 1.908 1.42l.261-.065a1.378 1.378 0 0 0 1.012-.965c.22-.816.533-2.512.062-4.51.136.02.285.037.443.051.713.065 1.669.071 2.516-.211.518-.173.994-.68 1.2-1.272a1.896 1.896 0 0 0-.234-1.734c.058-.118.103-.242.138-.362.077-.27.113-.568.113-.856 0-.29-.036-.586-.113-.857a2.094 2.094 0 0 0-.16-.403c.169-.387.107-.82-.003-1.149a3.162 3.162 0 0 0-.488-.9c.054-.153.076-.313.076-.465a1.86 1.86 0 0 0-.253-.912C13.1.757 12.437.28 11.5.28H8c-.605 0-1.07.08-1.466.217a4.823 4.823 0 0 0-.97.485l-.048.029c-.504.308-.999.61-2.068.723C2.682 1.815 2 2.434 2 3.279v4c0 .851.685 1.433 1.357 1.616.849.232 1.574.787 2.132 1.41.56.626.914 1.28 1.039 1.638.199.575.356 1.54.428 2.591z" />
+      </svg></button>
+            <p>Vote Score: {voteScore}</p>
+        </div>
+    );
+}
```
alicagatay, line 43: You could actually also divide this component into 2 smaller components such as `LikeButton` and `DislikeButton`, and then unite them in another components such as `LikeDislikeButtons`. This way, you could make your components way more simpler. Remember, it is always best practice for a component to have only 1 purpose.

alicagatay (changes requested):
Pretty good Dewayne. The design is really entertaining, and the functionality also works really well. There can be some improvements towards the simplicity of your codebase and your components, but other than those I believe you are ready to start into the next level.

PR #184, opened 2023-01-25
ZA2 - Gelson Celestino - Full-Stack-Project-Assessment

GelsonGerson:


SallyMcGrath (changes requested):
Hi @GelsonGerson  . There are 879 files in this PR. Which ones were you hoping mentors review? 😉

Would it be a good idea to open a PR only with the pertinent files? (Never git add . in a codebase)

PR #189, opened 2023-01-26
ZA2-Riyaaz Singh - Full-Stack-Project-Assessment

Riyaazct:


Dedekind561 (changes requested):
Hi there @Riyaazct,

Have you checked the number of files in this diff. We can't review that many! 😁 
Please make sure you remove the node_modules from your PRs so we can review your work.

Dedekind561:
I beg your pardon - I think you're actually _removing_ node_modules from the server directory. I'll check your code out now 😎

Riyaazct:
Hi @Dedekind561 

I have removed those extra files, i'm really not sure why it does not update the pr to reflect those changes, I literally tried every solution i could find online to remedy this. 

What it looks like to me now is that those files are showing in the P
Screenshot_2
R just as a reflection of the action taken on it as you can see by the screenshot I'm attaching herewith, as you can see there's a blocked out minus sign in red which i believe is an indication of the removal of the file.
Screenshot_1
Screenshot_3

as you can see the files are missing from the repo.

I'm open to trying something else to remedy this, if you have a solution. I really am.

The final option i have but with very little time on my hands to do so, is to delete the repo in its entirety, re-fork it and re-submit it.

Riyaazct:
Not a problem Sir... Thank you.

PR #197, opened 2023-01-27
ZA2-Jora Catalina_Alexandra

AlexJora:


Craques:
Hello, looked at your reducer and it looks good, you may want to removed node_modules from git to make your work easier to review.

You can start with this solution on stackoverflow https://stackoverflow.com/a/50675909/4642046.

AlexJora:
Hi Craques! I deleted node modules. Afterward, I attempted to remove deleted files but I created this chaos. 
Capture

Craques:
for your case you would need to run:

-  git rm -r --cached ./server/node_modules

then commit  and push those changes, this should fix your issue

AlexJora:
Hi! When I run: git rm -r --cached ./server/node_modules, I get: fatal: pathspec './server/node_modules' did not match any files

PR #202, opened 2023-01-27
Added Searchbar, diplay ther videos add the rating and delete buttons

FranklinPTorres:


client/src/components/Videos.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { useState } from "react";
+import data from "./exampleresponse.json";
+import VideoCards from "./VideoCards";
+function Videos() {
+  const [videos, setVideos] = useState(data);
```
alicagatay, line 5: It seems like you haven't added the feature of adding a video to the list yet.

alicagatay (changes requested):
Very well done Franklin. I can clearly see that you have finished almost all the requirements for level 100. The only missing feature is that, you haven't added a button to add a video to your videos list. Also complete this task, and you will be finishing level 100.

Nice work, well done!

PR #209, opened 2023-01-31
remove node_modules from client dir

Dedekind561:


Dedekind561:
Weird. I thought the previous PR removed the `node_modules` from the `server` dir. But the diff from shows the `node_modules` being added to the `client` dir. 🤔 And yet the `node_modules` have been removed from `server` dir now. Spooky 👻

textbook:
Without an appropriate `.gitignore` they'll all get added back again, I'm on mobile right now but suggest adding a root `.gitignore` containing:
```
node_modules/
```
which will cover both client and server.

Dedekind561:
I think there was an intention here to deliberately omit the `.gitignore` so trainees would figure out how to do this before project phase. 
I can add a github action to block any PRs where the diff is too big if we're agreed to leave the `.gitignore` out.
cc @SallyMcGrath

illicitonion (approved):

SallyMcGrath (approved):
approved for next pr to progress

PR #245, opened 2023-02-08
NW5 Manchester - Doris Siu - SQL - Week 1

Doris-Siu:


client/src/AddVideo.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default function AddVideo({ addVideo }) {
+  let [input, setInput] = useState({
+    title: "",
+    url: "",
+    rating: 0,
+  });
+
+  const handleChange = (e) => {
+    setInput({ ...input, [e.target.name]: e.target.value });
```
Craques, line 11: Looks good, just a minor suggestion, would it be possible to destructure `name` and `value` from e.target as variables?

client/src/AddVideo.js
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  const onSubmit = (e) => {
+    e.preventDefault();
+    input.id = uuidv4();
+    addVideo(input);
+  };
+
+  function uuidv4() {
```
Craques, line 20: maybe you can use the uuid library for this unless there is a valid reason to build your own random character generator 

https://www.npmjs.com/package/uuid if you are using npm
https://yarnpkg.com/package/uuidv4 uf you are using yarn

PR #258, opened 2023-02-27
Updated submit and delete hander function

khmdagal:


khmdagal:
Thanks Mitch. We came from a Long way 😂.

PR #271, opened 2023-05-15
London9-Hengameh-Mohammadkhani-Full stack-Assessment

HeniMKH:


client/src/App.css
```diff
@@ -1,3 +1,29 @@
...
+  position: absolute;
+  top:0;
+  width: 500px;
+}
+/* margin-top: 0;
+margin-bottom: 0.5rem;
+font-weight: 500;
+line-height: 1.2; */
+
+App-header {
```
jonnywyatt, line 19: This selector won't have any effect with a `.` at the start of the class name

client/src/App.css
```diff
@@ -1,3 +1,29 @@
 .App {
   text-align: center;
 }
+ 
+  .video-container {
+  display: grid;
+   grid-template-columns: repeat(3, 1fr);
```
jonnywyatt, line 7: Good use of grid!

client/src/App.js
```diff
@@ -1,11 +1,175 @@
...
+    }
+    return null; // Return null for no validation errors
+  };
+
+  const addClickHandler = async () => {
+    if (!title) {
+      alert("Please enter a title");
+      return;
+    }
+    validateUrl(url);
```
jonnywyatt, line 79: if the validation fails, you want to prevent the new video being sent to the server

client/src/App.js
```diff
@@ -1,11 +1,175 @@
...
+        "https://full-stack-assessment.onrender.com/video",
+        {
+          title: title.toLowerCase(),
+          url: url.split("watch?v=")[1],
+        }
+      );
+      setTitle("");
+      setUrl("");
+      alert("Video has been sent");
+      window.location.reload();
```
jonnywyatt, line 91: Ideally you'll use the response from the server to update the videos state so the videos list will re-render without needing a full page load.

client/src/App.js
```diff
@@ -1,11 +1,175 @@
...
+
+      <Container className="p-4">
+        <Row>
+          <Col md={4} className="mb-4">
+            <Button variant="link" onClick={openHandler}>
+              Add Video
+            </Button>
+            {isOpen && (
+              <>
+                <InputGroup className="mb-3">
```
jonnywyatt, line 111: It's good for accessibility to wrap these inputs in a `form` element and connect its `onSubmit` event to `addClickHandler` instead of the button `onClick`. Then users can submit the form by pressing enter while in an input; also screen reader users will get more useful information.

client/src/App.js
```diff
@@ -1,11 +1,175 @@
...
+                  <InputGroup.Text>Title</InputGroup.Text>
```
jonnywyatt, line 112: I don't think this associates the label with the input using `htmlFor`; better to use an normal HTML `label` element if React Bootstrap can't do this.

client/src/Video.jsx
```diff
@@ -0,0 +1,68 @@
...
+      });
+    } catch (err) {
+      console.log(err);
+    }
+  };
+
+  const handleDelete = (id) => {
+    axios.delete(`https://full-stack-assessment.onrender.com/video/${id}`);
+    alert("video has been deleted");
+    window.location.reload();
```
jonnywyatt, line 29: Try updating your videos list in React state, to avoid having to do a full page reload.

client/src/Video.jsx
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  return (
+    <div>
+      <h6 className="title">{info.title}</h6>
```
jonnywyatt, line 34: You've missed a few levels in the heading hierarchy.  You have a page heading (h1) so these should probably be h2

client/src/Video.jsx
```diff
@@ -0,0 +1,68 @@
...
+function Video({ info }) {
+  const [likeCount, setLikeCount] = useState(info.rating);
+
+  const handleLike = (status, id) => {
+    let newRate = likeCount;
+    if (status === "like") {
+      newRate = newRate + 1;
+      setLikeCount(newRate);
+    } else {
+      newRate = newRate - 1;
```
jonnywyatt, line 14: Could put a check in, if rating reaches zero then don't subtract from it

jonnywyatt (commented):
Well done for putting this together!
Running it locally, the app points to the server at `onrender.com` which threw 500 errors.

Using React Bootstrap has saved you writing some CSS, but has resulted in some accessiblity  problems eg associating form labels with inputs. I'd recommend writing plain HTML and CSS at this stage, even if the design looks less good, you'll learn more about the basics of styling and accessibility.
Good error handling on the client and server 👏 
After a server request returns to the client, ideally you'd update your list of videos without doing a full page reload - it's not an ideal user experience.

client/src/App.js
HeniMKH: I will

client/src/Video.jsx
HeniMKH: sorry ,silly mistake

PR #273, opened 2023-05-16
London 9 - Shayan Mahnam - Full Stack App is ready to Work :D

ShayanMahnam:
demo: https://cyf-shayanmahnam-video-recommendation.netlify.app/
server and DB are on Render which is so slow you have to wait for the server to wake up then you can use this app


client/src/components/TopBar.css
```diff
@@ -0,0 +1,91 @@
...
+}
+
+.input-top {
+  min-width: 350px;
+  height: 50px;
+  position: relative;
+  border-radius: 50px;
+  background: conic-gradient(from var(--r), #222 0%, #eee 10%, #222 20%);
+  border: none;
+  animation: rotating 3s linear infinite;
```
jonnywyatt, line 24: Nice animation 🤩

client/src/components/TopBar.jsx
```diff
@@ -0,0 +1,134 @@
+import React, { useState } from "react";
+import "./TopBar.css";
+
+// Validation utility function
+const isValidInput = (title, url) => {
```
jonnywyatt, line 5: 👍 Good UX that you're validating first in the browser, the user will get feedback faster than if they had to wait for a request to come back.

client/src/components/TopBar.jsx
```diff
@@ -0,0 +1,134 @@
...
+    } else if (url.includes("&list=")) {
+      videoCode = url.split("watch?v=")[1].split("&")[0];
+    } else if (url.includes("watch?v=")) {
+      videoCode = url.split("watch?v=")[1];
+    } else {
+      setErrorMessage("Invalid YouTube URL format.");
+      return;
+    }
+
+    if (cards.length > 0) {
```
jonnywyatt, line 55: Nice touch to check for duplicates 👍

client/src/components/TopBar.jsx
```diff
@@ -0,0 +1,134 @@
...
+        return;
+      }
+    }
+
+    const newCard = {
+      title: title.trim(),
+      url: videoCode,
+    };
+
+    fetch("https://video-server-iiqf.onrender.com/videos", {
```
jonnywyatt, line 69: When running the client and server locally, it'd be even better if you could feed some .env variables into the app to point the client to localhost if running locally, and to onrender if running in 'production' mode.

client/src/index.css
```diff
@@ -1,10 +1,15 @@
 body {
   margin: 0;
+  padding: 0;
+  box-sizing: border-box;
```
jonnywyatt, line 4: Could apply these to all elements with a reset:
```
* {
	box-sizing: border-box;
	...(reset other properties eg padding, margin)
}
```

client/yarn.lock
jonnywyatt, line 1: Best to use either NPM or Yarn but not both, so that all users of the repo get a consistent result.

jonnywyatt (commented):
You've nailed this; very clear well-organised code (your code is self-documenting with clear variable names so don't feel obliged to add code comments).
You could push on even further by looking at ways to organise CSS / reduce duplication - eg SASS / CSS modules / CSS variables.
Bravo 👏

client/yarn.lock
ShayanMahnam: yes, that’s why I deleted this. I don’t why in the initial repo we forked from cyf they had both yarn and npm

client/src/components/TopBar.css
```diff
@@ -0,0 +1,91 @@
...
+  transition: opacity 0.3s ease;
+  border: none;
+  color: white;
+  font-size: 20px;
+  padding-left: 1rem;
+}
+
+textarea:focus,
+input:focus {
+  outline: none;
```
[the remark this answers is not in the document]
ShayanMahnam: This is so cool thank you

client/src/components/TopBar.jsx
ShayanMahnam: can you please explain more about this?

jonnywyatt: The line above has a hardcoded UR for the videos endpoint on onrender.com.
If you're running the code locally, it would be good to be able to point the app to the locally running server (on localhost) rather than the server on onrender.com.
Ideally you can do this without changing the code each time. 
Your app code could make a decision to point to either localhost or onrender.com using techniques like:
- change the app code to check the browser hostname - if it's localhost then make requests to the server on localhost. If the browser hostname is netlify, or wherever the app is deployed to, then make requests to the server at onrender.com
- use a .env file locally to load some environment variables into the app. You can read them with a library like dotenv. The .env file would be in your gitignore, so not in git, and would not be present when the app is running on Netlify. Then, your default could be to use the server at onrender.com, unless an environment variable is loaded that overrides that and points to the server on localhost

ShayanMahnam: I added env. thank you so much

PR #285, opened 2023-05-18
LONDON_9-SAGHAR_HOSSEINMARDI_FULL-STACK-PROJECT-ASSESSMENT

SagharHosseinmardi:
Demo: https://full-stack-project-assessment-client-7ywt.onrender.com

### DIFF
--- a/client/src/Input.jsx
+++ b/client/src/Input.jsx
@@ -1,0 +1,64 @@
+import React, { useState } from "react";
+
+function Input({ onAddVideo }) {
+  const [title, setTitle] = useState("");
+  const [url, setUrl] = useState("");
+
+  const handleSubmit = (e) => {
+  e.preventDefault();
+  if (isValidYouTubeURL(url) && isValidTitle(title)) {
+    const embedUrl = generateEmbedURL(url);
+    onAddVideo(title, embedUrl);
+    setTitle("");
+    setUrl("");
+  } else {
+    alert("Please enter a valid YouTube URL and a non-empty title.");
+  }
+  };
+  
+  const isValidTitle = (title) => {
+    // Trim the title to remove leading and trailing spaces
+    const trimmedTitle = title.trim();
+    return trimmedTitle.length > 0;
+  };
+
+
+  const isValidYouTubeURL = (url) => {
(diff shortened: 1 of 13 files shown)


client/src/Input.jsx
```diff
@@ -0,0 +1,64 @@
+import React, { useState } from "react";
+
+function Input({ onAddVideo }) {
```
jonnywyatt, line 3: The component name is a bit confusing, something like `AddVideoForm` could be a bit clearer?

client/src/Input.jsx
```diff
@@ -0,0 +1,64 @@
...
+    const embedUrl = generateEmbedURL(url);
+    onAddVideo(title, embedUrl);
+    setTitle("");
+    setUrl("");
+  } else {
+    alert("Please enter a valid YouTube URL and a non-empty title.");
+  }
+  };
+  
+  const isValidTitle = (title) => {
```
jonnywyatt, line 19: It's good User Experience that you're validating in the front end app rather than sending the request to the back end and only allowing that to validate. This way, the user gets feedback quicker, without waiting for the request to go to the server and come back. In a production app, you'd probably validate on the server as well just for security purposes, in case someone sent a request directly to it.

server/server.js
```diff
@@ -1,15 +1,150 @@
...
 app.get("/", (req, res) => {
   // Delete this line after you've confirmed your server is running
   res.send({ express: "Your Backend Service is Running" });
 });
+
+app.get("/videos", async (req, res) => {
+  try {
+    const query = "SELECT * FROM videos";
+    const result = await pool.query(query); //seconde query can be anything
+    const orderedVideos = result.rows;
```
jonnywyatt, line 94: It doesn't look like a sort order is being applied yet, so this variable name is a bit confusing.

jonnywyatt (commented):
Well done Saghar, you've kept things simple but you're doing all the fundamentals well here.
On the front end, markup structure and semantics are solid eg heading hierarchy, accessible forms, use of buttons.
On the back end, simple readable code with error handling.
If you're confident with the code you've written, this is a great foundation to move forward from.

PR #286, opened 2023-05-19
London9-Mohamed Abdi-Full-Stack-Project-Assessment

MohamedAbdi114:
@JDysiewicz


client/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
JDysiewicz, line 1: Nice, thanks for putting this in - makes the PR much easier to see :+1:

client/src/App.js
```diff
@@ -1,11 +1,64 @@
+import React, { useState } from "react";
 import "./App.css";
+import data from "./exampleresponse.json";
+import VideoList from "./VideoList";
+import AddVideo from "./AddVideo";
+import SearchVideo from "./SearchVideo";
 
 function App() {
+  const [VideosData, setVideosData] = useState(data);
```
JDysiewicz, line 9: Careful with the naming here; JS variables are all `lowerCamelCase`, `UpperCamelCase` is only for `class`es.

client/src/App.js
```diff
@@ -1,11 +1,64 @@
...
+  const [searchVideos, setSearchVideos] = useState([]);
+  console.log(VideosData);
+
+  function addVideo(title, url) {
+    let newVideo = {
+      id: Math.floor(Math.random() * 1000000),
```
JDysiewicz, line 15: Better to use a dedicated ID generating package (e.g. `uuid`), but if you're replacing this with a full-stack version (e.g. where you let the database handle the ID) then it's fine as a stop-gap.

client/src/App.js
```diff
@@ -1,11 +1,64 @@
...
       </header>
+      <div className="body-section">
+        <SearchVideo search={search} />
+        <AddVideo addVideo={addVideo} />
+
+        {searchVideos.length > 0 ? (
+          <VideoList data={searchVideos} delete={removeVideo} />
+        ) : (
+          <VideoList data={VideosData} delete={removeVideo} />
+        )}
```
JDysiewicz, line 60: You're returning the same thing regardless of length here - I assume you mean to return `null` when there's no `searchVideo.length`?

client/src/Votes.js
```diff
@@ -0,0 +1,28 @@
+import React, { useState } from "react";
+
+function Votes() {
+  const [Like, setLike] = useState(0);
+
+  const handleLike = () => {
+    setLike(Votes + 1);
```
JDysiewicz, line 7: I think this is a typo; this is meant to be `setLike(like+1)`? Another way to do this is `setLike((currentValue) => currentValue+1)`

JDysiewicz (commented):
Looking pretty good so far, a few typos to fix up to get it working smoothly, but it's all making sense so far and the structure of the code looks fine :)

MohamedAbdi114:
@JDysiewicz  can you review my level 200 & 250 please .thank you

JDysiewicz (commented):
Looks good, couple small changes

MohamedAbdi114:
@JDysiewicz could you please view my Level 300 as well . thank you

client/src/App.js
```diff
@@ -14,7 +15,8 @@ function App() {
     fetch("http://127.0.0.1:5000/")
```
JDysiewicz, line 15: This works, but use `localhost` instead of `127.0.0.1` as it's more readable

server/.env
```diff
@@ -0,0 +1,136 @@
+USERNAME= mohamedabdi
```
JDysiewicz, line 1: This `.env` file should be in the `.gitignore` file - this has commited secret values to version control

server/db.js
```diff
@@ -0,0 +1,11 @@
+const Pool = require("pg").Pool;
+
+const pool = new Pool({
+  user: "postgres",
+  host: "localhost",
+  database: "cyf_videos",
+  password: "101",
+  port: 5432,
+});
```
JDysiewicz, line 9: This shoulod be pulled from environment variables instead

server/server.js
```diff
@@ -71,32 +74,101 @@ let videos = [
...
+// app.get("/", (req, res) => {
+// Delete this line after you've confirmed your server is running
+//   res.send(videos);
+// });
+
 // GET "/"
-app.get("/", (req, res) => {
-  // Delete this line after you've confirmed your server is running
-  res.send(videos);
+app.get("/videos", async (req, res) => {
```
JDysiewicz, line 84: Nice use of `async/await`

server/server.js
```diff
@@ -71,32 +74,101 @@ let videos = [
...
+        .status(400)
+        .json({ result: "failure", message: "Title and URL are required" });
+    }
+    const newVideo = await pool.query(
+      "INSERT INTO VIDEOS (title,url,rating) VALUES($1, $2, $3) RETURNING *",
+      [title, url, 0]
+    );
+    res.json(newVideo.rows[0]);
+  } catch (err) {
+    console.log(err.message);
```
JDysiewicz, line 114: If you're handling the error, make sure to return a value to the user `res.status(x).json({data: null, error: err})` or something like that

JDysiewicz (commented):
A couple comments, but the server-side code looks really nice. You've got a nice pattern going on and it's looking nice, just make sure to handle all those errors with response codes, and careful not to commit database secrets to version control.

PR #296, opened 2023-05-24
London9 - Bahare Bahreinian - Full-Stack-Project-Assessment - Week-3

Bahare09:


client/src/components/VCard.js
```diff
@@ -0,0 +1,33 @@
...
+    onDelete(videoId);
+  };
+  const handleRatingUpdate = (videoId, newRating) => {
+    onUpdateRating(videoId, newRating);
+  };
+  return (
+    <div className="video-container">
+      {videoData.map((video) => (
+        <div key={video.id} className="video-card">
+          <p>{video.title}</p>
```
jonnywyatt, line 15: As this is a subheading, an `h2` would be better

client/src/index.css
```diff
@@ -1,13 +1,158 @@
...
+}
+
+.dataLoading {
+  font-size: 1.3rem;
+}
+.addVideo {
+  margin: 1rem 0;
+}
+
+.addVideo label {
```
jonnywyatt, line 69: I don't think this is used?

client/src/components/Rating.js
```diff
@@ -0,0 +1,30 @@
...
+  const handleClickDownVote = () => {
+    let newRating = 0;
+    vote > 0 ? (newRating = vote - 1) : (newRating = 0);
+    setVote(newRating);
+    onUpdateRating(newRating);
+  };
+
+  return (
+    <div className="rating">
+      <FaThumbsUp className="icon-2" onClick={handleClickUpVote} />
```
jonnywyatt, line 23: It's better to put these icons inside `button` elements. Eg a keyboard user will be able to tab to them.

client/src/components/Footer.js
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+function Footer() {
+  return (
+    <footer className="footer">
+      <h3>Designed by Bahare Bahreinian</h3>
```
jonnywyatt, line 6: This could probably just be a `p` as there's no content under it, so it's not a subheading

client/src/index.css
```diff
@@ -1,13 +1,158 @@
...
+}
+
+.footer h3 {
+  color: blanchedalmond;
+  padding: 1.5rem;
+  font-size: 1rem;
+  text-align: center;
+}
+
+@media screen and (max-width: 1240px) {
```
jonnywyatt, line 142: This works fine, but better to switch it round so that the default styles (ie not in a media query) are for small screens, then use media query(s) with `min-width` to update the styles for wider screens. This is called 'mobile first' and is the most common industry practice.

client/src/components/HomePage.js
```diff
@@ -0,0 +1,100 @@
...
+  const updateVideoData = (newVideoData) => {
+    fetch(`https://video-3.onrender.com/?order=${sortOrder}`, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify(newVideoData),
+    })
+      .then((response) => response.json())
+      .then((data) => setVideoData(data))
```
jonnywyatt, line 35: Small point, but the server could send back a 400. If `response.ok` is false, you know there's an error so you shouldn't call `setVideoData`

jonnywyatt (commented):
Great job!
There are lots of good things in this PR including:
- Clean, small React components
- Semantic choices of HTML elements like `button`, `header`, `footer`
- Good use of flex for CSS layouts
- Error handling for promises eg fetch
- Use of async / await
- Error handling and HTTP error codes in the server code

When developing locally, it'd be useful to be able to point the front end app at the server running on `localhost:5000` using env variables instead of having to edit the fetch URLs in HomePage.js.

Well done 👏

client/src/components/HomePage.js
```diff
@@ -0,0 +1,100 @@
...
+        setIsLoading(false);
+      })
+      .catch((error) => {
+        console.log(error);
+        setIsLoading(false);
+      });
+  }, []);
+
+  const updateVideoData = (newVideoData) => {
+    fetch(`https://video-3.onrender.com/?order=${sortOrder}`, {
```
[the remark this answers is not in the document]
jonnywyatt: @Bahare09 has correctly pointed out that this is used for sort order in the server code.

client/src/components/HomePage.js
```diff
@@ -0,0 +1,100 @@
...
+      },
+      body: JSON.stringify(newVideoData),
+    })
+      .then((response) => response.json())
+      .then((data) => setVideoData(data))
+      .catch((error) => console.log(error));
+  };
+
+  const deleteVideo = (videoId) => {
+    fetch(`https://video-3.onrender.com/${videoId}/?order=${sortOrder}`, {
```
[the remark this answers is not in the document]
jonnywyatt: @Bahare09 has correctly pointed out that this is used for sort order in the server code.

PR #301, opened 2023-05-26
London 9 - Karleen Richards - Full Stack Project Assessment

karleenmsrichards:


karleenmsrichards:
> Hi @karleenmsrichards I tried to review your PR but when I run `npm install` I get dependency conflicts. If you run `npm install` on your PR branch locally, are you seeing any errors?

Hi hi, I  tried what you asked and I do not get any errors installing npm, it says that it is up to date and there are no issues.

karleenmsrichards:
> Hi @karleenmsrichards I tried to review your PR but when I run `npm install` I get dependency conflicts. If you run `npm install` on your PR branch locally, are you seeing any errors?

Can you try running npm install --legacy-peer-deps, I think I did have some problems installing it with just npm install.

client/src/About.css
```diff
@@ -0,0 +1,31 @@
+.about-card-wrapper {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    text-align: center;
+    padding: 40px 0 60px 0;
+    gap: 10px;
+}
+  
+@media only screen and (min-width: 375px) {
```
jonnywyatt, line 10: Great, mobile first 👏 
Although 375px as a breakpoint will include almost every device out there, it's more common to introduce breakpoints at higher width to adapt the layout. 
You don't need to repeat all the properties from the default selector above, just the ones that have changed (in this case, looks like padding and margin only)

client/src/About.css
```diff
@@ -0,0 +1,31 @@
...
+      text-align: center;
+      padding: 100px 0 20px 0;
+      gap: 10px;
+      margin: 220px auto 60px auto;
+    }
+}
+
+.about-link {
+    padding-top: 10px;
+    color: #e94de9;
```
jonnywyatt, line 24: If you find you re-use colours, CSS variables can be a good way to give them readable names so you don't have to remember which hex value is which colour

client/src/AddVideoForm.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  useEffect(() => {}, [videos]);
+
+  return (
+    <ScrollAnimation animateIn="animate__animated animate__backInLeft">
+      <form onSubmit={handleSubmit} className="home-form">
+        <legend>Add your video here</legend>
+        <div className="form-section-wrapper">
+          <label htmlFor="video-url" className="form-label">
```
jonnywyatt, line 62: Great for accessibility eg screen readers that you're associating the labels with the inputs with `htmlFor`

client/src/AddVideoForm.js
```diff
@@ -0,0 +1,91 @@
...
+      })
+      .catch((error) => alert(error.message));
+  }
+
+  useEffect(() => {}, [videos]);
+
+  return (
+    <ScrollAnimation animateIn="animate__animated animate__backInLeft">
+      <form onSubmit={handleSubmit} className="home-form">
+        <legend>Add your video here</legend>
```
jonnywyatt, line 60: `legend` is for use with `fieldset`, you could either wrap the fields in one, or just use a heading tag for the form title

client/src/AddVideoForm.js
```diff
@@ -0,0 +1,91 @@
...
+      //   return response.json();
+      // })
+
+      .then((response) => {
+        if (response.status > 400 && response.status < 500) {
+          throw new Error(
+            "Please ensure your video has a title and a valid Youtube url."
+          );
+        }
+        if (response.status > 500 && response.status < 600) {
```
jonnywyatt, line 42: It doesn't look like you're handling 500 status, could use `>== 500`

client/src/AddVideoForm.js
```diff
@@ -0,0 +1,91 @@
...
+      })
+      .then((newVideo) => {
+        setVideos((value) => [...value, newVideo]);
+        setVideoTitle("");
+        setVideoUrl("");
+      })
+      .catch((error) => alert(error.message));
+  }
+
+  useEffect(() => {}, [videos]);
```
jonnywyatt, line 55: Not needed?

client/src/AllVideos.css
```diff
@@ -0,0 +1,199 @@
+* {
```
jonnywyatt, line 1: I'd put a reset like this in the top level stylesheet eg App.css, you wouldn't normally find resets in component stylesheets

client/src/AllVideos.css
```diff
@@ -0,0 +1,199 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: 20px;
+}
+
+@media only screen and (min-width: 375px) {
+    .all-videos-wrapper {
+        width: 100%;
+        margin: 120px 0;
+        list-style: none;
```
jonnywyatt, line 20: I don't think this is needed, it's not a list element

client/src/AllVideos.css
```diff
@@ -0,0 +1,199 @@
...
+    padding-left: 100px;
+} 
+
+.all-videos-filter-btn {
+    background-color: transparent;
+    color: white;
+    border: none;
+}
+
+@media only screen and (min-width: 375px) {
```
jonnywyatt, line 54: It doesn't look like these properties are being changed, can remove the media query?

client/src/AllVideos.css
```diff
@@ -0,0 +1,199 @@
...
+    row-gap: 40px;
+    max-width: 100%;
+    padding: 0 20px 10px 20px;
+}
+
+@media only screen and (min-width: 375px) {
+    .all-videos-container {
+        padding: 0 80px;
+        display: grid;
+        grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));
```
jonnywyatt, line 96: Great!

client/src/App.js
```diff
@@ -1,11 +1,75 @@
...
+import { AboutPage } from "./AboutPage";
+
+export let videosContext = createContext(null);
 
 function App() {
+  const [videos, setVideos] = useState([]);
+  const [isDesc, setIsDesc] = useState(true);
+
+  const url = "https://full-stack-project-server.onrender.com/videos";
+  const fetchData = useCallback(() => {
```
jonnywyatt, line 22: It could be just me, and it's a clever use of `useCallback` but this is a little confusing to understand; would it be simpler to just extract `fetchData` to outside the component so it's not redeclared on each render? Then the dependencies for the `useEffect` could just become `[url, isDesc]`.
But it's just my opinion, this is the kind of pattern worth discussing with other devs as well.

client/src/Card.js
```diff
@@ -0,0 +1,23 @@
...
+  cardTextClass,
+  cardText,
+  cardLink,
+  cardLinkClass,
+  cardLinkText,
+}) => {
+  return (
+    <div className={wrapperClass}>
+      <p className="card-subtitle">{cardSubtitle}</p>
+      <h1 className="card-title">{cardTitle}</h1>
```
jonnywyatt, line 16: There should only be one `h1` per page, so this should probably be a lower level heading?

client/src/Header.css
```diff
@@ -0,0 +1,100 @@
+* {
```
jonnywyatt, line 1: No need to repeat resets, just have one copy in the top level stylesheet.

client/src/HomeHero.js
```diff
@@ -0,0 +1,24 @@
...
+        <p className="home-hero-text">
+          Cats are graceful, agile and playful, and they make wonderful
+          companions. They are among the most loved and cherished pets in the
+          world. To watch more videos please{" "}
+          <a href="#all-videos" className="home-hero-all-videos-link">
+            click here
+          </a>
+          .
+        </p>
+        <a href="#addvideo" className="add-video-link">
```
jonnywyatt, line 18: This is javascript only functionality not a link, so should be a `button` element

client/src/Login.js
```diff
@@ -0,0 +1,81 @@
+import "./Login.css";
```
jonnywyatt, line 1: I can't see where login is connected to the running app, sorry if I missed it!

client/src/Login.js
```diff
@@ -0,0 +1,81 @@
...
+      <form onSubmit={handleLogin} className="login-form">
+        <legend>Login to your account</legend>
+        <label htmlFor="email" className="login-form-label">
+          Email
+        </label>
+        <input
+          type="email"
+          id="email"
+          onChange={(e) => setEmail(e.target.value)}
+          value={email}
```
jonnywyatt, line 58: You could set the input type to `email`, will add useful keys eg to mobile keyboards

client/src/Login.js
```diff
@@ -0,0 +1,81 @@
...
+          Password
+        </label>
+        <input
+          type="password"
+          id="password"
+          onChange={(e) => setPassword(e.target.value)}
+          value={password}
+          className="form-input"
+        />
+        <button onClick={register} className="register-btn">
```
jonnywyatt, line 72: Ideally if the user hits enter while in a field, the form submits. Trying to cover both register and login submissions in one form is tricky, probably easier to have two forms.

server/.env
```diff
@@ -0,0 +1,11 @@
+# DB_HOST=db.tchaegbogyejdhrymbni.supabase.co
```
jonnywyatt, line 1: This file should be in your .gitignore, so secrets like passwords don't get committed to git

server/server.js
```diff
@@ -1,15 +1,135 @@
...
+    const allVideos = await videosPool.query(query);
+    res.json(allVideos.rows);
+  } catch (error) {
+    console.log(error);
+    res.status(500).send(error.message);
+  }
+});
+
+app.post("/videos/:id/like", async (req, res) => {
+  const videoId = req.body.id;
```
jonnywyatt, line 39: You can also get it from `req.params.id` so not essential to send it in the body

jonnywyatt (commented):
Amazing ambition to produce a slick interface and animations.
Some very good practice as well to use mobile first media queries (although 375px is very low as a breakpoint, almost every device / width will be caught be it; also no need to repeat all the properties that won't change), and accessibility. 
Great use of flex and grid.
As the amount of CSS you write increases, you could use SASS / CSS variables to manage colours and spacing units, and extract common code to mixins.
Using React context for the video list was a good solution to avoid drilling those as props through layers of the component hierarchy.
Solid error handling in the Node.js code.
Bravo 👏

client/src/About.css
karleenmsrichards: Have made changes so that only the necessary is changed in the media query. Thanks for pointing that out. About the media query and the min-width, ideally, I intend to add another breakpoint at 768px and probably max-width 1400px in future so this was a start tocapture a mobile version for the least. But, I think you are saying that I can start higher for instance at min-width 540px. Is that figure more common, or what would you recommend please?

client/src/About.css
karleenmsrichards: Yes, it's true. I have used them in other projects. I will consider adding them to this one as well. It totally makes sense right. 👍

client/src/AddVideoForm.js
karleenmsrichards: Good spot. The changes were made. Thank you.

client/src/AddVideoForm.js
karleenmsrichards: Thanks for the reminder. I wrapped the legend and the rest of the form in a fieldset.

client/src/AllVideos.css
karleenmsrichards: Oh I see, if I do have it at the top level, would I then need to import the App.css for instance to this component?

client/src/AllVideos.css
karleenmsrichards: Just checked, It doesn't appear to be needed as you suggested, thank you.

client/src/AllVideos.css
karleenmsrichards: Sure thing.

client/src/App.js
karleenmsrichards: Can you explain to me what you mean in a different way please, I'm not sure if I understood what you meant?

client/src/HomeHero.js
karleenmsrichards: I have changed it to a button element. Thanks much.

client/src/Login.js
karleenmsrichards: No it's not connected, I'm not sure if I would continue with the log in bit.

client/src/Login.js
karleenmsrichards: I think it is currently set to type="email".

client/src/Login.js
karleenmsrichards: Oh I see, noted.

server/server.js
karleenmsrichards: Yes, it is true, will make the necessary changes. thank you very much. 😺

karleenmsrichards:
@jonnywyatt Immensely grateful for your feedback. I have understood almost all you said apart from 2 questions I have about the min-width media query suggestions  and extracting the fetchData function.  I can make those changes once I have a better insight. I left the questions where you commented on those. I have made all other changes. thank you very much for your support and encouragement.

client/src/About.css
jonnywyatt: Advice on media queries can differ; in some case breakpoints can be added when the layout requires it, ie when it starts to break or not work so well.
An alternative is to set breakpoints at 'classic' device widths - a very common one is 768px (iPad portrait width). Some advice here - https://css-tricks.com/snippets/css/media-queries-for-standard-devices/

client/src/AllVideos.css
jonnywyatt: No, as long as the top level stylesheet is loaded into the page somewhere, it doesn't need to be directly imported into this stylesheet. All the selectors will be considered together.

client/src/App.js
jonnywyatt: Maybe I'm reading it wrong but it looks like the use of useCallback is to avoid fetchData being redeclared on re-renders, and the prop passed to AllVideos therefore changing and causing it to re-render. It makes sense and is good practice, but fetchData could just be declared once outside the component (as it probably would be in a production app, in it's own file and unit-tested). That will achieve the same thing and is a bit simpler.

PR #308, opened 2023-05-31
London Class 9 - Farzaneh Haghani - Full-Stack-Project-Assessment

farzaneh-haghani:
https://farzaneh-haghani-video-recommendation.netlify.app


client/src/components/Add.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const addHandler = async (event) => {
+    event.preventDefault();
+    if (
+      title === "" ||
+      url === "" ||
+      title.trim() === "" ||
+      url.trim() === ""
+    ) {
+      alert("Please Fill All Fields!");
```
JDysiewicz, line 18: Might be a nice touch to specify what fields need filling out; e.g. make an array of all the fields which don't satisfy the condition and display that to the user.

server/server.js
```diff
@@ -1,15 +1,141 @@
...
+
+
+
 const app = express();
-const port = process.env.PORT || 5000;
 
+app.use(express.json());
+app.use(cors());
+
+const port = process.env.PORT || 5000;
```
JDysiewicz, line 12: Try and use `??` over `||` in most cases; `||` triggers on falsey values which is often not the intention when it's used.

server/server.js
```diff
@@ -1,15 +1,141 @@
...
-  res.send({ express: "Your Backend Service is Running" });
+  res.status(200).json({ data: videos });
 });
+
+
+// // POST "/add"
+app.post("/add", (req, res) => {
+  const { title, url } = req.body;
+  const regex1 = /(^https:\/\/www\.youtube\.com\/watch\?v=)([0-9a-zA-Z=\-_+.,:;'?!@#|$%^&*<>()]){11}$/;
+  const regex2 = /(^https:\/\/www.youtube.com\/shorts\/)([0-9a-zA-Z=\-_+.,:;'?!@#|$%^*<>()]){11}$/;
```
JDysiewicz, line 90: When using regex, it's good to add a comment explaining what the regex does because regex is horrible to read even when written well.

server/server.js
```diff
@@ -1,15 +1,141 @@
...
+  const result1 = regex1.test(url);
+  const result2 = regex2.test(url);
+  if (!result1 && !result2) {
+    res.status(400).json({
+      "result": "failure",
+      "message": "URL is not valid!"
+    });
+  }
+  else {
+    const id = Math.floor(Math.random() * 1000000);
```
JDysiewicz, line 100: Don't rely on random number generation to create IDs; it's fine in a toy application but look into using a UUID generator instead.

server/server.js
```diff
@@ -1,15 +1,141 @@
...
+      "message": "Video could not be found!"
+    })
+  else
+    res.status(200).json({ data: result })
+})
+
+
+// DELETE "/"
+app.delete("/:id", (req, res) => {
+  const id = parseInt(req.params.id);
```
JDysiewicz, line 127: This can throw an error if id is not parseable as a number (e.g. `/not-a-number` would make `id = NaN` which will cause issues further down). It's good to add a `if` statement checking that id is not `NaN` (note, there's a function to do this, don't do `id != NaN` as JavaScript doesn't like it for historic reasons.

JDysiewicz (commented):
Application looks super clean, and similar to what I'd expect in a production code base. Really good job. A few things to clean up, but the overall codebase looks nice. Start having a think about how you want to organise files in your codebase; you've got `/components` , but `Home, Add, Search` are all complete pages in your application (as you have a router which links directly to them), therefore having these in a `/pages` directory then having reuseable component in `/components` would probably be a bit nicer.

server/server.js
farzaneh-haghani: I did it for temporary because the next step is database and I will make it to generate serial id.

farzaneh-haghani:
@sampennington  I finished
#Needs code Review

PR #315, opened 2023-06-05
London 9 - Vitalina Kuzmenko - Full Stack Project Assessment

VitalinaKuzmenko:
https://videovortex-vitalina-kuzmenko.netlify.app/


server/server.js
```diff
@@ -1,15 +1,152 @@
...
+
+function generateUniqueId() {
+  return Math.floor(Math.random() * 1000000) + 1;
+}
+
+// Error handling middleware
+app.use((err, req, res, next) => {
+  console.error("Error:", err);
+  res.status(500).json({
+    result: "failure",
```
sampennington, line 24: Oh well this is cool! Didnt know you could do that

server/server.js
```diff
@@ -1,15 +1,152 @@
...
+
+    if (action === "thumbs-up") {
+      video.rating = parseInt(video.rating) + 1;
+    } else if (action === "thumbs-down") {
+      video.rating = parseInt(video.rating) - 1;
+    }
+
+    const updateQuery = "UPDATE videos SET rating = $1 WHERE id = $2";
+    await db.query(updateQuery, [video.rating, id]);
+
```
sampennington, line 93: Nice handling of various http status codes etc 🚀

server/server.js
```diff
@@ -1,15 +1,152 @@
...
+// DELETE "/{id}"
+app.delete("/:id", async (req, res, next) => {
+  const { id } = req.params;
+
+  try {
+    const query = "DELETE FROM videos WHERE id = $1";
+    await db.query(query, [id]);
+
+    res.json({});
```
sampennington, line 146: Should we respond with a 200 here maybe, or does it do it anyway? 🤔 
(...i dont use express much 🙈 )

sampennington (commented):
Looks great  Vitalina 😄

server/server.js
VitalinaKuzmenko: Thank you! Google helped me to know about it :D

server/server.js
VitalinaKuzmenko: It implies a 200 status code by default. It's a common practice to use res.json({}) for empty responses in REST APIs.
After I delete the video I fetch all videos (so I use another endpoint) to update the list on page. That's why I didn't send anything as a response to front end but only an empty array :)

PR #323, opened 2023-08-01
Update docs to refer to PostgreSQL and Render

sztupy:
Heroku is no longer free, and the trainees are now taught PostgreSQL. The option to use Mongo is still in the notes if they want to, but the docs will link to the PostgreSQL setup for Render


Dedekind561 (approved):
LGTM, 
thankyou for updating this!

PR #334, opened 2023-08-15
London10_Olha-Danylevska_Full-Stack-Project

OlhaDanylevska:


ShadiSafaee (commented):
Hi @OlhaDanylevska  -  In server could you add node_modules to .gitignore file? You can follow [this ](https://medium.com/@rspesc31/forgot-to-add-node-modules-to-gitignore-solved-6c5b7a444c65)to find out how to do so. Currently, I'm unable to view your code because your node_modules have consumed a significant amount of storage in your pull request.

OlhaDanylevska:
hi @ShadiSafaee, yes, I did it wrong last time, gitignore was not in root directory, seems like now everything is fine

client/src/Counter.js
```diff
@@ -0,0 +1,69 @@
+import { useState } from "react"
+const Counter = ({ setAllMyVideos, videoId, videoRating }) => {
+    const [rating, setRating] = useState(0)
+
+    const handleCounterDown = () => {
+
+        if (rating <= 0) {
+            setRating(0)
+            handleRatingChange(rating)
```
illicitonion, line 9: `setRating` doesn't actually update `rating` right away, it schedules it to be updated in the future - if you `console.log` it you should see it still has the old value - you may want to set a variable you can use here, e.g.
```javascript
const newRating = rating <= 0 ? 0 : rating - 1;
handleRatingChange(newRating);
```

You may see at the moment that your app is one click behind on rating updates :)

server/server.js
```diff
@@ -1,15 +1,158 @@
...
+// });
+
+
+app.post("/videos", (req, res) => {
+  let newVideoTitle = req.body.title
+  let newVideoURL = req.body.url
+  let newVideoRating = 0
+
+  if (!newVideoTitle || !newVideoURL || newVideoTitle === "" || newVideoURL === "") {
+    res.status(400).send("Fill in the missing fields")
```
illicitonion, line 89: It can be useful in an error message to say _which_ fields were missing, so that it can be displayed to the user and they know what to fix :)

illicitonion (commented):
This is generally looking solid, congrats!

client/src/Counter.js
OlhaDanylevska: Updated counter, working a little bit better, not perfect, but better

client/src/Counter.js
```diff
@@ -0,0 +1,72 @@
...
+            newRating = 0
+            setRating(newRating)
+            handleRatingChange(newRating)
+            getAllNewVideos()
+        } else {
+            newRating = rating - 1
+            setRating(newRating)
+            handleRatingChange(newRating)
+            getAllNewVideos()
+        }
```
illicitonion, line 16: Your two branches here mostly do the same thing - I would factor out the common things outside of the branch:
```suggestion
        if (rating <= 0) {
            newRating = 0
        } else {
            newRating = rating - 1
        }
        setRating(newRating)
        handleRatingChange(newRating)
        getAllNewVideos()
```

There are two reasons for this:
1. If we need to change what we do, we will only need to change it in one place (and can't forget to change the other)
2. It makes it more clear to the reader of the code what's _actually_ different, and what's the same, in the two cases.

client/src/Counter.js
```diff
@@ -0,0 +1,72 @@
...
+    const handleCounterUp = () => {
+        newRating = rating + 1
+        setRating(newRating)
+        handleRatingChange(newRating)
+        getAllNewVideos()
+    }
+
+
+
+    const handleRatingChange = async (newRating) => {
```
illicitonion, line 27: I would maybe call this function `changeRating` or `pushNewRating` rather than `handleRatingChange` - this function is the thing that _actually changes the rating on the server_ - "handle" as a prefix for a function name generally indicates something that happens _in response to_ the rest of the name (i.e. I would assume that this function is called _after the rating is changed in response to it_ rather than _actually changing the rating itself_).

client/src/Counter.js
```diff
@@ -0,0 +1,72 @@
...
+            const response = await fetch(`https://youtube-video-server.onrender.com/videos/${videoId}`, {
+                method: "PUT",
+                headers: {
+                    "Content-Type": "application/json",
+                },
+                body: JSON.stringify({ rating: newRating }),
+            });
+
+            if (!response.ok) {
+                console.error("Failed to update rating");
```
illicitonion, line 39: Real users don't see errors in the console :) If this were a real project, I would recommend two things:
1. Showing a message to the user saying something went wrong
2. Resetting the rating state back to the previous value before you tried to change it, so that it doesn't look like the change took effect.

There's a spectrum of options to consider between "update the UI instantly and ignore any errors", "update the UI but reset it if something went wrong", or "don't update the UI (but maybe show a loading spinner or something) until the server confirms the update was successful" - when do you think each is the best approach?

server/server.js
```diff
@@ -1,15 +1,162 @@
...
+  let newVideoTitle = req.body.title
+  let newVideoURL = req.body.url
+  let newVideoRating = 0
+
+  if (!newVideoTitle) {
+    res.status(400).send("Title field is missing");
+  } else if (!newVideoURL) {
+    res.status(400).send("URL field is missing");
+  } else if (!newVideoTitle && !newVideoURL) {
+    res.status(400).send("Both title and URL fields are missing");
```
illicitonion, line 93: Is it possible for this code to ever run? What situation would lead to this code running?

server/videos-full-stack.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE videos 
+    (
+        id SERIAL PRIMARY KEY,
+        title VARCHAR(50) NOT NULL,
```
illicitonion, line 4: You only allow titles of up to 50 characters in your database schema - what would happen if someone tried to post a new video with a longer title? What would you like/expect to happen?

illicitonion (commented):
This is looking really nicely put together, congratulations! I have a few style comments to get you thinking, but the functionality all looks great!

PR #337, opened 2023-08-18
London 10 | Maksim Lukianenko | Fullstack Project Assignment | Level 999

Mpanasetckiy:
I completed the project using JS, React, Node, and MongoDB via Mongoose. While building the project, I utilised components from previous projects, such as the Modal window and custom http-hook. Also for better consistency and security reasons, I created .env files to store API key and other data in both front and back ends without pushing them online.


client/src/App.js
```diff
@@ -1,13 +1,43 @@
 import "./App.css";
+import React, { useState } from "react";
+
+import UploadComponent from "./components/UploadComponent";
+import Search from "./components/Search";
+import Container from "./components/Container";
+import Footer from "./components/Footer";
+
+const App = () => {
+  const [data, setData] = useState([]);
```
illicitonion, line 10: "data" is a very generic name for a piece of state - can you think of a name that better describes _what_ data is stored here?

client/src/App.js
```diff
@@ -1,13 +1,43 @@
...
+  const [videos, setVideos] = useState([]);
+  const [searchQuery, setSearchQuery] = useState("");
+
+  const addLocally = (newData) => {
+    data.push(newData);
+  };
+
+  const searchForVideo = () => {
+    const filteredVideos = data.filter((video) => {
+      return video.title.toLowerCase().includes(searchQuery);
```
illicitonion, line 20: Is the search query guaranteed to be lower case? If not, what will happen if I search for "Hello"?

client/src/App.js
```diff
@@ -1,13 +1,43 @@
...
+    });
+    setVideos(filteredVideos);
```
illicitonion, line 22: I would generally expect you not to discard the original list of videos, because ideally you wouldn't need to re-fetch them if I remove/change my search query... Can you think how to avoid throwing away the full list of videos when searching?

client/src/components/Clip.jsx
```diff
@@ -0,0 +1,73 @@
...
+  };
+
+  const handleHover = () => {
+    setCurrentClip({ id: id, rating: rating });
+  };
+
+  const changeRating = ({ target }) => {
+    if (target.className === "like-button") {
+      setCurrentClip({ ...currentClip, rating: currentClip.rating + 1 });
+      updateRating(currentClip.id, currentClip.rating + 1);
```
illicitonion, line 36: It's interesting to need both of these calls - I would kind of expect just one function to be passed as a prop to update rating, and that would update all of the things needed - what do you think?

client/src/components/UploadComponent.jsx
```diff
@@ -0,0 +1,87 @@
+import React, { useState } from "react";
+import { v4 as uuidv4 } from "uuid";
+
+import { useHttpClient } from "../hooks/http-hook";
+
+const UploadComponent = ({ addVideo }) => {
+  const randomId = uuidv4();
+  const [videoData, setVideoData] = useState({
+    id: randomId,
```
illicitonion, line 9: I'd generally expect the server to be generating IDs (because it can guarantee they're actually unique), rather than trusting the client to give it a unique value.

client/src/components/UploadComponent.jsx
```diff
@@ -0,0 +1,87 @@
...
+      addVideo(videoData);
+      setVideoData({
+        id: randomId,
+        url: "",
+        title: "",
+        rating: 0,
+      });
+      console.log(response);
+    } catch (err) {
+      console.log("Error adding video.", { error: err });
```
illicitonion, line 43: A user will never see this error - how can you ensure the user actually knows when something went wrong, and what they should do about it?

illicitonion (commented):
This generally looks really nicely put together, good job!

PR #356, opened 2023-08-24
London 10 | Andrius Isin | Full-Stack-Project-Assessment | Level 300

AndriusIsin:
FRONTEND: https://video-recomendations.onrender.com
BACKEND SERVER: https://video-recomendations-server.onrender.com/


client/src/Components/OneVideoCard.js
```diff
@@ -0,0 +1,43 @@
+import { useState } from "react";
+const OneVideoCard = ({ id, title, url, rating, handleDelete }) => {
+  const [videoRating, setVideoRating] = useState(rating);
+
+  const increase = () => {
+    setVideoRating(videoRating + 1);
```
kitko112, line 6: FYI, setVideoRating(rating => rating + 1) also works, this approach it wont need the videoRating var on line number 3 as the set function can get that from the param. both approaches work in this case

client/src/App.css
```diff
@@ -1,3 +1,27 @@
-.App {
+
+body::before {
+  display: block;
+  content: '';
```
kitko112, line 4: Is this css content necessary?

client/src/App.js
```diff
@@ -1,12 +1,61 @@
...
-    <div className="App">
-      <header className="App-header">
-        <h1>Video Recommendation</h1>
-      </header>
-    </div>
+    <>
+      <NavBar />
+      <Welcome />
+      <AddAndSearch setKeyword={setKeyword} />
+      <AddVideo setVideoCards={setVideoCards} videoCards={videoCards} />
```
kitko112, line 44: The logic works, one suggestion that helps improve the readability, we can write the if, else if and else in this way so that it is easier to read

```
if(isLoading){
   return (
     <h1 className="blink_me">Loading.....</h1>
)

if(error) {
  return ( <h1 className="error-msg "> {error}</h1>)
}
  // main logic
return (
     <VideoList ... />
)
}
```

client/src/Components/OneVideoCard.js
AndriusIsin: Thanks for the review :)

client/src/App.js
AndriusIsin: Thanks, I'll make this change.

client/src/Components/AddVideo.js
```diff
@@ -0,0 +1,115 @@
...
+        body: JSON.stringify(newVideo),
+      })
+        .then((res) => res.json())
+        .then((data) => {
+          setVideoCards(data);
+          setRebuild((rebuild) => rebuild + 1);
+        })
+        .catch((error) => console.log(error));
+    } else {
+      alert("Please fill fields correctly");
```
illicitonion, line 30: It can be more useful friendly to tell them which fields were incorrect, and what was expected, e.g. if I'd put a non-youtube link as the URL, telling me that you expect a youtube link would be helpful :)

client/src/Components/OneVideoCard.js
```diff
@@ -0,0 +1,63 @@
...
+  const decrease = () => {
+    if (videoRating > 0) {
+      setVideoRating(videoRating - 1);
+    }
+    fetch(`http://localhost:4000/videos/${id}`, {
+      method: "PUT",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ rating: videoRating - 1 }),
```
illicitonion, line 27: Interesting - in one of these you cap the value at 0, but in the other you let the value go below zero... Is that intentional?

client/src/Components/VideoList.js
```diff
@@ -0,0 +1,53 @@
...
+      console.log(error.message);
+    }
+  };
+  return (
+    <>
+      <section className="p-5" id="recommendations">
+        <div className="container">
+          <div className="row row-cols-auto text-center custom-gap ">
+            {videoCards
+              .filter((video) => {
```
illicitonion, line 28: Filter accepts a callback function which returns a truthy value - I'd ordinarily expect the callback function to return `true`/`false` rather than `video`/`undefined`...

client/src/Components/VideoList.js
```diff
@@ -0,0 +1,53 @@
...
+                  return video;
+                }
+              })
+              .map((video) => (
+                <OneVideoCard
+                  key={video.id}
+                  id={video.id}
+                  title={video.title}
+                  rating={video.rating}
+                  url={video.url.substring(video.url.indexOf("=") + 1)}
```
illicitonion, line 43: What happens if the video url doesn't contain `=`?

illicitonion (commented):
Looking really good! 🎉

AndriusIsin:
> Looking really good! 🎉

Thanks, Daniel, for your review :)  I have made changes

PR #360, opened 2023-08-24
London_10-Saliha_Popal/ Full-Stack-Project-Assessment/ Feature level 250

SalihaPopal:


client/src/AddVideoForm.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+    const validateUrl = (url) => {
+    // Simple validation for YouTube URLs
+    return /^https?:\/\/(www\.)?youtube\.com\/watch\?v=[\w-]+$/.test(url);
+  };
+
+  const handleAddVideo = () => {
+
+    if (title.trim() === '') {
+      setErrorMessage('Please enter a title');
```
illicitonion, line 25: Good job making sure the user knows the problem and what to do about it!

client/src/AddVideoForm.jsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    const newVideo = {
+      id: Date.now(), // Generate a unique ID
+      title: title,
+      url: url,
+      rating: video.rating,
+    };
+    onAddVideo(newVideo);
+    setTitle('');
```
illicitonion, line 41: This looks very similar to the code in `handleCancel` - could you maybe rename that function to reflect what it _does_ rather than when you call it, and also call it here?

client/src/App.js
```diff
@@ -1,11 +1,39 @@
...
     <div className="App">
       <header className="App-header">
-        <h1>Video Recommendation</h1>
+        <h4>Video Recommendation</h4>
       </header>
+      {videos.map(video => (
+      <div key={video.id}>
+        <h5>{video.title}</h5>
+        <p>{video.url}</p>
+        <p>{`rating: video.rating`}</p>
```
illicitonion, line 33: What does this show in the page? Is it what you expect?

client/src/VideoItem.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+
+const VideoItem = ({ video, onDelete, onVote }) => {
+
+  const videoId = video.url.split('v=')[1];
```
illicitonion, line 5: What would happen if the video URL didn't contain `v=`? Can you think of a way of handling that more smoothly?

server/server.js
```diff
@@ -1,15 +1,56 @@
...
+});
+
+// POST A VIDEO TO THE API
+app.post('/:id', (req, res) => {
+  console.log(req.body, " <---- Add a new video to the API");
+  const {title, url} = req.body;
+  if(!title || !url){
+    response.status(404).json({
+      "result": "failure",
+      "message": "Video could not be saved"
```
illicitonion, line 49: Could you include in the error message what was wrong with the request, so it could be better reported to the user what went wrong?

illicitonion (commented):
This looks like it should work pretty well - good job!

PR #365, opened 2023-08-24
London10 | Seyyednavid Hejazijouybari | Full-Stack-Project-Assessment | Level 999

seyyednavid:


client/src/components/Video.js
```diff
@@ -0,0 +1,112 @@
...
+  const deleteHandler = async () => {
+    try {
+      const response = await fetch(`${process.env.REACT_APP_SERVERURL}/${id}`, {
+        method: "DELETE",
+      });
+      if (response.status !== 200) {
+        throw new Error("Something went wrong!");
+      }
+      const data = await response.json();
+      getDeleteMessage(data.message);
```
illicitonion, line 24: It's not super clear from the name what `getDeleteMessage` does - could you think of a more clear name to show what this means?

client/src/components/Video.js
```diff
@@ -0,0 +1,112 @@
...
+      if (response.status !== 200) {
+        throw new Error("Something went wrong!");
+      }
+      getAllVideos();
+    } catch (error) {
+      console.error(error);
+    }
+  };
+
+  const voteDownHandler = async () => {
```
illicitonion, line 48: This looks very similar to the `voteUpHandler` - can you work out how to share their code in one function, rather than duplicate it?

client/src/components/VideoForm.js
```diff
@@ -0,0 +1,108 @@
+import { useState } from "react";
+const VideoForm = ({ getAllVideos }) => {
+  const [title, setTitle] = useState("");
+  const [url, setUrl] = useState("");
+  const [backendMessage, setBackendMessage] = useState("");
```
illicitonion, line 5: What's the backend message? Can you think of a more clear name for this, or add a comment explaining what it's for?

client/src/components/VideoForm.js
```diff
@@ -0,0 +1,108 @@
...
+        body: JSON.stringify(newVideo),
+      });
+      if (response.status === 400) {
+        const errorData = await response.json();
+        const validationErrors = errorData.errors;
+        setErrorMessage(validationErrors.map((error) => error.msg).join(", "));
+      } else if (response.status === 409) {
+        const existingError = await response.json();
+        setErrorMessage(existingError.message);
+      } else if (response.status !== 201) {
```
illicitonion, line 42: Given you've been doing lots of `===` checks before, I'd more expect you to handle the `=== 201` case here, and for the else to be "all other status codes", rather than the other way around. Your core here works, but it's kind of surprising, which makes it slightly harder to read and work with. People are really good at spotting patterns, so breaking that pattern can be confusing.

client/src/pages/SearchPage.js
```diff
@@ -0,0 +1,94 @@
+import { useEffect, useState } from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faSearch } from "@fortawesome/free-solid-svg-icons";
+import VideoLists from "../components/VideoLists";
+const SearchPage = ({ allVideos, getAllVideos, order }) => {
+  const [search, setSearch] = useState("");
+  const [videoSearch, setVideoSearch] = useState([]);
```
illicitonion, line 7: It's not super clear from the naming what `videoSearch` is and how it's different from `search` - can you think of clearer names to make their differences and roles clear?

Or is there a way you could even get rid of one of the states, and not need them both?

client/src/pages/SearchPage.js
```diff
@@ -0,0 +1,94 @@
...
+      setVideoSearch([]);
+    } else {
+      setSearch(searchText);
+      setSelectedValue("");
+    }
+  };
+
+  useEffect(() => {
+    if (search === "") {
+      return;
```
illicitonion, line 24: If you've already searched for something, then get rid of the search term, what will be displayed?

server/db.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE videos_recommendation;
+
+CREATE TABLE videos (
+    id VARCHAR(255) PRIMARY KEY,
+    Date VARCHAR(255),
```
illicitonion, line 5: Postgres has specific date types that are more suitable to use than `VARCHAR(255)`, I'd recommend using one as it will allow you to do things like comparisons, or extracting a month, more easily in the future :)

server/db.sql
```diff
@@ -0,0 +1,9 @@
+CREATE DATABASE videos_recommendation;
+
+CREATE TABLE videos (
+    id VARCHAR(255) PRIMARY KEY,
```
illicitonion, line 4: We typically use `SERIAL` as a type of ids, rather than `VARCHAR`s - can you research `SERIAL` and work out what it does and why it's useful?

server/server.js
```diff
@@ -1,15 +1,170 @@
...
-app.listen(port, () => console.log(`Listening on port ${port}`));
+// Get all videos
+app.get("/", async (req, res) => {
+  const { order, search } = req.query;
+  try {
+    const allVideos = await pool.query("SELECT * FROM videos");
+    let filteredVideos = [...allVideos.rows];
+    // Apply search filter if search query is provided
+    if (search) {
+      filteredVideos = filteredVideos.filter((video) =>
```
illicitonion, line 20: Can you do this filtering in SQL rather than in javascript? Also the sorting.

Both work, but doing so in SQL will be faster.

server/server.js
```diff
@@ -1,15 +1,170 @@
...
+        video.title.toLowerCase().includes(search.toLowerCase())
+      );
+    }
+    //Apply order filter
+    if (order === "desc") {
```
illicitonion, line 25: I think you're sorting on both the client side and the server side - do you need to do both? What are the advantages/disadvantages of doing it in each place?

illicitonion (commented):
This is generally looking really good, but I'd encourage you to spend more time thinking about making sure the names you use are really clear and meaningful.

PR #372, opened 2023-08-26
London 10 - Paulina Wywrot - Full Stack Assessment - Level 300

PaulinaWywrot:


client/src/Components/AddVideo.js
```diff
@@ -0,0 +1,62 @@
...
+  const [videoLink, setVideoLink] = useState("");
+  function handleInputChange(event) {
+    if (event.target.name === "name") {
+      setVideoName(event.target.value);
+    } else if (event.target.name === "url") {
+      setVideoLink(event.target.value);
+    }
+  }
+
+  function handleSubmit(event) {
```
ShadiSafaee, line 14: When you log videos after updating it, you may notice that it doesn't immediately reflect the new state because `setState` is `asynchronous`. If you want to see the updated state, you can log it inside a useEffect hook with videos as a dependency. Might worth trying and compare it with useState 😃

client/src/Components/AllVideos.js
```diff
@@ -0,0 +1,15 @@
+import VideoCard from "./VideoCard";
+
+const AllVideos = ({ videos, setVideos }) => {
+  const removeVideo = (video) =>
+    setVideos(videos.filter((oneVideo) => oneVideo !== video));
+  return (
+    <div className="row">
```
ShadiSafaee, line 7: Your code looks good to me! 
But you can improve it further by using `semantic HTML` and making it more accessible. You can use an unordered list to represent a list of videos, and each video can be represented as a list item. `ul` and `li` tags instead of `div`

client/src/Components/VideoCard.js
```diff
@@ -0,0 +1,50 @@
+import { useState } from "react";
+
+const VideoCard = ({ video, removeVideo }) => {
+  const [rating, setRating] = useState(video.rating);
+
+  const addVote = () => {
+    setRating(rating + 1);
+    console.log(rating);
```
ShadiSafaee, line 8: Would be good not to push `console.log` into remote! remember other devs wouldn't need to see lines you added for debugging.

client/src/Components/VideoCard.js
```diff
@@ -0,0 +1,50 @@
...
+        <button className="btn btn-outline-success" onClick={addVote}>
+          Add Vote
+        </button>
+        <h4>Rating: {rating}</h4>
+        <button className="btn btn-outline-warning" onClick={deleteVote}>
+          Delete Vote
+        </button>
+      </div>
+      <button
+        class="btn btn-danger btn-lg btn-block mt-2 p-1"
```
ShadiSafaee, line 39: You can update `class` to `className` for consistency. Seems that in React, it's conventional to use the `className` attribute instead of `class` for specifying CSS classes

client/src/Components/VideoCard.js
```diff
@@ -0,0 +1,50 @@
...
+  const addVote = () => {
+    setRating(rating + 1);
+    console.log(rating);
+  };
+  const deleteVote = () => {
+    setRating(rating - 1);
+    console.log(rating);
+  };
+
+  let url = video.url.substring(video.url.indexOf("=") + 1);
```
ShadiSafaee, line 15: Ah nice way of parsing the YouTube video ID from the URL!🔥 
The only thing it might not work for all YouTube `urls`.

server/server.js
```diff
@@ -1,15 +1,116 @@
...
+    id: id,
+    title: req.body.title,
+    url: req.body.url,
+    rating: Math.floor(Math.random() * 100),
+  };
+  if (req.body.title && req.body.url) {
+    videos.push(newVideo);
+    res.json(videos);
+  } else
+    res.status(404).send({
```
ShadiSafaee, line 98: HTTP status code `404` typically shows that the request was not found, which may not be the most appropriate status code for this situation. Maybe (400, Bad Request) instead of 404 suits better!

ShadiSafaee (commented):
Hi @PaulinaWywrot - very well-done, great job so far! 
A general suggestion would be finding out where you could use `useEffect` in your React app . Good luck with the next levels! 🥇

client/src/Components/AddVideo.js
```diff
@@ -0,0 +1,72 @@
...
+        },
+        body: JSON.stringify(video),
+      })
+        .then((res) => res.json())
+        .then((data) => {
+          setVideos(data);
+        })
+        .catch((error) => console.log(error));
+    } else {
+      alert("Please ensure fields are filled in correctly");
```
illicitonion, line 34: In general if you're telling a user they've done something wrong, telling them what's expected is useful - I'd maybe change this to something like:
```suggestion
      alert(`Expected video URL to start with 'https://www.youtube.com/' but you provided '${video.url}' - please ensure fields are filled in correctly`);
```

client/src/Components/VideoCard.js
```diff
@@ -0,0 +1,48 @@
+import { useState } from "react";
+
+const VideoCard = ({ video, removeVideo }) => {
+  const [rating, setRating] = useState(video.rating);
```
illicitonion, line 4: It looks like these ratings aren't stored in the database, so if you refresh the page they'll go away, and if multiple users look at the page they won't see each other's ratings. How would you add these to the database so that they're persisted?

illicitonion (commented):
This is looking great, nicely put together!

server/server.js
```diff
@@ -1,15 +1,143 @@
...
-app.get("/", (req, res) => {
-  // Delete this line after you've confirmed your server is running
-  res.send({ express: "Your Backend Service is Running" });
+app.get("/videos", async (req, res) => {
+  try {
+    const result = await db.query("SELECT * FROM videos");
+    res.status(200).json(result.rows);
+  } catch (error) {
+    console.log(error);
+    res.status(404).json({ error: "failed to fetch videos" });
```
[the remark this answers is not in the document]
illicitonion: (Consider the status codes you use for all responses, too - is there a more appropriate one?)

PaulinaWywrot:
> Hi @PaulinaWywrot - very well-done, great job so far! A general suggestion would be finding out where you could use `useEffect` in your React app . Good luck with the next levels! 🥇

Thank you for your comments Shadi, I've made some changes to my code :)

PaulinaWywrot:
> This is looking great, nicely put together!


Hello @illicitonion, thank you for your review. I've implemented some changes as suggested, but please let me know if you have any further suggestions:)

PR #375, opened 2023-08-26
london-10-shadi-fakhri-full-stack-project

Shadi38:


server/server.js
```diff
@@ -1,15 +1,229 @@
...
 
-// GET "/"
-app.get("/", (req, res) => {
-  // Delete this line after you've confirmed your server is running
-  res.send({ express: "Your Backend Service is Running" });
+// GET "/videos"
+app.get("/videos", async function (req, res) {
+  const result = await db.query("SELECT * FROM videos");
+  if (result.rows.length === 0) {
+    return res.status(404).json({ error: "no videos found" });
```
illicitonion, line 45: It's pretty common to only 404 if we're looking for one specific thing that couldn't be found, and if you're asking for "all the videos", to just return the empty array and let the client handle that. What you've done here isn't wrong, it's totally valid, just raising that there are alternatives too :) Can you think of some advantages/disadvantages of both approaches?

server/server.js
```diff
@@ -1,15 +1,229 @@
...
+//     orderVideos = videos.sort((a, b) => a.rating - b.rating);
+//   } else {
+//     // Default to descending order if order is not specified or is invalid
+//     orderVideos = videos.sort((a, b) => b.rating - a.rating);
+//   }
+
+//   return res.json(orderVideos);
+// });
+//ordering by assending and dessending for example /videos?order=asc
+app.get("/videos", async function (req, res) {
```
illicitonion, line 71: Why do you have a second handler function for the same method on the same path? Which one will end up getting used?

server/server.js
```diff
@@ -1,15 +1,229 @@
...
+  const result = await db.query("SELECT * FROM videos");
```
illicitonion, line 72: Can you do the sorting in SQL rather than in javascript? (Both work, but doing it in SQL will probably be a bit faster :))

server/server.js
```diff
@@ -1,15 +1,229 @@
...
+        res.status(201).json(createdVideo);
+      }
+    });
+  }
+);
+
+//search video by query for example /videos/search?title=halleluja
+app.get("/videos/search", function (req, res) {
+  const searchVideo = req.query.title;
+  const filteredVideo = videos.filter((video) =>
```
illicitonion, line 153: This doesn't use your database - can you rewrite this to do a SQL query (and like with the sorting, to do the filtering in SQL not JS? Again - the JS still works, but doing it in SQL will probably be faster).

server/server.js
```diff
@@ -1,15 +1,229 @@
...
+  const result = await db.query('SELECT * FROM videos where id=$1',[videoId]);
+  if (result.rows.length === 0) {
+    return res.status(404).json({ error: "no videos found" });
+  }
+  res.json(result.rows);
+});
+
+//updating video with id
+app.put("/videos/:id", function (req, res) {
+  const video = videos.find((m) => m.id == req.params.id);
```
illicitonion, line 185: Can you do this in SQL? :)

illicitonion (commented):
This is looking really good, congrats!

client/src/App.js
```diff
@@ -1,12 +1,92 @@
...
+        const data = await response.json();
+        //asending acording to the rating
+        data.sort((a, b) => a.rating - b.rating);
+        return setLoadData(data);
+      } catch (error) {
+        console.error("Error fetching data:", error);
+      }
+  }
+
+async function descClickHandler(e) {
```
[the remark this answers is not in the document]
Shadi38: you are right I fixed it.

client/src/App.js
```diff
@@ -1,12 +1,92 @@
...
+  e.preventDefault();
+  try {
+    const response = await fetch("https://web-server-5nme.onrender.com/videos");
```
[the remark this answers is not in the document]
Shadi38: I deleted fetch from the function.

client/src/Videos.js
```diff
@@ -0,0 +1,277 @@
+import React, { useEffect, useState } from "react";
+import ThumbUpIcon from "@mui/icons-material/ThumbUp";
+import ThumbDownIcon from "@mui/icons-material/ThumbDown";
+function Videos(props) {
+  const {loadData,setLoadData}=props;
+  // const [loadData, setLoadData] = useState([]);
+  const [titleData, setTitleData] = useState("");
+  const [urlData, setUrlData] = useState("");
```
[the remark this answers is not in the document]
Shadi38: I changed them to title and url

server/server.js
Shadi38: I did it in Sql

client/src/Videos.js
```diff
@@ -0,0 +1,277 @@
+import React, { useEffect, useState } from "react";
+import ThumbUpIcon from "@mui/icons-material/ThumbUp";
+import ThumbDownIcon from "@mui/icons-material/ThumbDown";
+function Videos(props) {
+  const {loadData,setLoadData}=props;
```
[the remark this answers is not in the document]
Shadi38: I changed it.

server/server.js
Shadi38: thanks for the good advice. I changed that.

server/server.js
Shadi38: I did it.

Shadi38:
thanks, Daniel for reviewing my project. your advice was really good and helped my project to get better. I tried to do all of them. I appreciate it if you could check my commits and let me know your opinion of them.

PR #379, opened 2023-08-27
London-10_Saqib-Javed_full-stack_level_100_200_300

saqibjvd:
level 100 WIP


client/src/Components/NewVideo.js
```diff
@@ -0,0 +1,60 @@
+import React, { useState } from "react";
+
+function NewVideo(props) {
+  // const { newVideos, handleNewVideo, handleSubmit} = props
+
+  const [title, setTitle] = useState("");
+  const [url, setUrl] = useState("");
+  const [newVideo, SetNewVideo] = useState({ title: "", url: "" });
```
ShadiSafaee, line 8: Nice use of destructuring to initialize the newVideo state!

client/src/Components/NewVideo.js
```diff
@@ -0,0 +1,60 @@
...
+      // console.log("updated value", updatedValue);
+      props.onAddNewVideo(updatedValue)
+      setTitle("")
+      setUrl("")
+    }
+  };
+
+  return (
+    <div className="new-video">
+      <p>Add new video to your list</p>
```
ShadiSafaee, line 29: Using an `<h3>` element is also a good choice for the title or heading of a section.

client/src/App.js
```diff
@@ -1,13 +1,24 @@
...
+import Videos from "./Components/Videos";
+
 
 function App() {
+
   return (
     <div className="App">
       <header className="App-header">
         <h1>Video Recommendation</h1>
+        <div>
```
ShadiSafaee, line 13: FYI Components do not need to be wrapped in a `<div>`

client/src/Components/VideoCard.js
```diff
@@ -0,0 +1,50 @@
...
+
+  function dislike() {
+    setRating(rating - 1);
+  }
+
+    return (
+      <div>
+        
+        <div></div>
+        <div key={video.id}>
```
ShadiSafaee, line 21: The key prop should be applied to the outermost element that is part of a list of elements. In your case, it should be applied to the outer `<div>` that wraps each video card, not the inner div. 
FYI this is important for React to efficiently update the component list.

client/src/Components/Videos.js
```diff
@@ -0,0 +1,46 @@
...
+    setVideos((prevVideosData) =>
+      prevVideosData.filter((video) => video.id !== id)
+    );
+  };
+
+  const addVideos = (video) => {
+    console.log("this is video", video);
+    setVideos([video, ...videos]);
+    console.log("in add videos", videos);
+  };
```
ShadiSafaee, line 19: Good job in useing `spread operator`! can you also remove unused codes in your PR? `console.log`s

server/server.js
```diff
@@ -1,15 +1,162 @@
...
+    if (!newVideo.title || !newVideo.url || !newVideo.url.startsWith("https://www.youtube.com")) {
+      res.status(404).send({
+        result: "failure",
+        message: "Video could not be saved",
+      });
+    } else {
+      const videoid = videos.map((video) => video.id);
+      const id = Math.max(...videoid) + 1;
+      
+      videos.push(newVideo);
```
ShadiSafaee, line 110: when adding a new video, you can also consider what happens if `videos.push(newVideo)` fails :)

server/server.js
```diff
@@ -1,15 +1,162 @@
...
+    res.status(400).send("No matching video with this ID exists.");
+  } else {
+    res.status(200).json({ matchingVideo });
+  }
+});
+
+//Deletes the video with the ID container within the {id} parameter
+
+app.delete("/videos/:id", (req, res) => {
+  let id = Number(req.params.id);
```
ShadiSafaee, line 140: Since `id` in your code doesn't need to be reassigned, you can use `const` instead.

client/src/Components/VideoCard.js
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  function dislike() {
+    setRating(rating - 1);
+  }
+
+    return (
+      <div>
+        
+        <div></div>
```
ShadiSafaee, line 20: Also you don't need this line I guess.

server/server.js
```diff
@@ -1,15 +1,162 @@
...
+
+// Returns the video with the ID contained within the {id} parameter.
+
+app.get("/videos/:id", (req, res) => {
+  const id = Number(req.params.id);
+  const matchingVideo = videos.find((video) => {
+    return video.id === id;
+  });
+  if (!matchingVideo) {
+    res.status(400).send("No matching video with this ID exists.");
```
ShadiSafaee, line 131: Instead of sending a 400 status code for "No matching video with this ID exists," you might consider using a 404 status code which is more suitable here

ShadiSafaee (commented):
Great job @saqibjvd! 
I've just spotted very minor issues and good luck with DB step and deployment!

client/src/Components/NewVideo.js
illicitonion, line 8: What is this state for? I don't see it being set or read anywhere?

server/server.js
```diff
@@ -1,15 +1,162 @@
...
+      id: 0,
+      title: req.body.title,
+      url: req.body.url,
+      rating: 0
+    };
+
+    if (!newVideo.title || !newVideo.url || !newVideo.url.startsWith("https://www.youtube.com")) {
+      res.status(404).send({
+        result: "failure",
+        message: "Video could not be saved",
```
illicitonion, line 104: It can be useful to include what the specific problem was in this message, so it can be displayed to the user (e.g. that it was specifically the title that was missing, or that the URL wasn't a YouTube URL, etc)

illicitonion (commented):
This is generally looking good, congrats!

PR #380, opened 2023-08-27
London 10 -- Kristina Dudnyk -- Fullstack Project Assignment -- DataBases -- 300

KristinaDudnyk:


GergelyKI (commented):
Looks good all together, I left one comment.

client/build/asset-manifest.json
```diff
@@ -0,0 +1,13 @@
...
+    "main.js": "/static/js/main.100d7049.js",
+    "index.html": "/index.html",
+    "main.964fdded.css.map": "/static/css/main.964fdded.css.map",
+    "main.100d7049.js.map": "/static/js/main.100d7049.js.map"
+  },
+  "entrypoints": [
+    "static/css/main.964fdded.css",
+    "static/js/main.100d7049.js"
+  ]
+}
```
GergelyKI, line 13: In general it's a good idea too add the build folder to the git ignore file, so that these bit's don't clatter your source code.

client/src/NewVideo.js
```diff
@@ -0,0 +1,73 @@
+import { useState } from "react";
+
+export default function NewVivdeo({ setRefreshVideos }) {
+  const [formData, setFormData] = useState({
```
illicitonion, line 4: It's good to group these related pieces of data together, nicely done!

(Though I'm not sure about `id` and `rating`? They seem to be unused in this component?)

client/src/VideosPerent.js
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from "react";
+import VideoCard from "./VideoCard";
+
+export default function VideosPerent({ refreshVideos, setRefreshVideos }) {
+  const [videos, setVideos] = useState([]);
+  const [rating, setRating] = useState([]);
```
illicitonion, line 6: Interesting - this `rating` state has a singular name (rating not ratings), but contains an array... And then when you use it, you iterate over each of the videos, so you're considering one particular video, but you pass the whole array of ratings as a prop to the `VideoCard`... This is a little surprising - I would generally expect the rating to be part of the video object (or, in unusual situations, maybe to have two arrays which have the same number of elements in them, and to iterate across them together in one loop), but this usage pattern is a little weird... Are you sure about it? How did you end up here? Happy to discuss! :)

client/src/SetRating.js
```diff
@@ -0,0 +1,13 @@
+import { useState } from "react";
+
+export default function SetRating({ ratin, setRating, video }) {
```
illicitonion, line 3: This component doesn't seem to use its `ratin` or `setRating` props? Why does it have them?

server/server.js
```diff
@@ -1,15 +1,73 @@
...
+app.use(express.static("public"));
+app.use(express.json());
+app.use(express.urlencoded({ extended: true }));
 
 // Store and retrieve your videos from here
 // If you want, you can copy "exampleresponse.json" into here to have some data to work with
-let videos = [];
+// let videos = [];
+
+function getConvertedDataFromBookingsJSON() {
```
illicitonion, line 21: Bookings? :)

server/server.js
```diff
@@ -1,15 +1,73 @@
...
+app.post("/video", (req, res) => {
+  const { title, url } = req.body;
+  if (!title && !url) {
+    res.status(400).json({ error: "One or several filds wasn't filed" });
+  }
+
+  const objData = getConvertedDataFromBookingsJSON();
+  objData.push({
+    ///////////////////////////////////////////////
+    id: req.body.url.split("=")[1],
```
illicitonion, line 61: What if the URL didn't contain an `=`? What would happen then? How can we handle that?

illicitonion (commented):
This is generally looking pretty good, but it'd be worth thinking/talking about how to handle related pieces of state (e.g. your ratings state array). Also, did you get on to the database parts yet?

server/server.js
KristinaDudnyk: Good point, I have fixed that in my new commit. 
When a user makes a PUSH request, it sends the corrected URL controlled inside the handleSubmit, which makes sure that it's passing the fixed format of URL.

server/server.js
KristinaDudnyk: It was the reference 😖

client/src/VideosPerent.js
KristinaDudnyk: Doing coding in the middle of the night wasn't a great idea.
I rewrote everything on the next morning╰(*°▽°*)╯

client/src/SetRating.js
KristinaDudnyk: Right!!
Thanks

client/src/NewVideo.js
KristinaDudnyk: Thanks, I have minimalized passing formData

client/src/NewVideo.jsx
```diff
@@ -0,0 +1,89 @@
...
+    if (formData.title.trim() === "" || formData.url.trim() === "") {
+      alert("Title and URL cannot be empty");
+      return;
+    }
+
+    const formDataUrlChanged = {
+      title: formData.title,
+      url: `https://www.youtube-nocookie.com/embed/${
+        formData.url.match(/v=([a-zA-Z0-9_-]{11})/)[1]
+      }`,
```
illicitonion, line 23: Can you talk through what this code is doing?

What assumptions are you making, and how confident are you in them? What could go wrong with it if your assumptions are violated?

client/src/UpdateRating.jsx
```diff
@@ -0,0 +1,45 @@
+import { useState } from "react";
+
+export default function UpdateRating({ video }) {
+  const [counter, setCounter] = useState(video.rating);
+
+  const handleUpdate = async () => {
+    try {
+      const response = await fetch(
+        `https://kristinadudnyk-fullstack-project.onrender.com/video/${video.id}`,
```
illicitonion, line 9: I've noticed you've put in this URL a few times, and that in each of them you have a localhost:4500 version commented out.

Can you think of a way to make it easier for you to change which server your frontend is talking to for when you want to test things locally?

server/server.js
```diff
@@ -1,15 +1,129 @@
...
+  } catch (err) {
+    console.error(err);
+    res.status(500).json({ message: "Internal server error" });
+  }
+});
+
+app.post("/video", async (req, res) => {
+  const { title, url } = req.body;
+  // console.log("post video req.body", req.body);
+  const id = url.split("embed/")[1];
```
illicitonion, line 58: What assumptions are you making about input from users here? What would happen if those violations were violated? How could they be?

illicitonion (commented):
Really nice clean-ups, good job! 🎉

client/src/NewVideo.jsx
KristinaDudnyk: It's a regex that returns the first match, which is eleven characters after v= sign 
There was a thought that link could have params and queries, depending on user's action.
So that code is preventing this possibility from the very beginning.

client/src/UpdateRating.jsx
KristinaDudnyk: A brainy mate suggested that I can use 
```
ssl: {
    rejectUnauthorized: false,
  },
```
which makes it run locally

server/server.js
KristinaDudnyk: Regex changes it before the request is sent, so it prevents that issue on the front-end side, passing the clean URL to the server.

client/build/asset-manifest.json
KristinaDudnyk: Thank you!

PR #382, opened 2023-08-28
Bekir London 10 FullStackProject

BekirKuscu:


BEKIRKSU:
All 300 levels done.

ShadiSafaee (commented):
Hi @BEKIRKSU - could you add node_modules to .gitignore file? You can follow [this ](https://medium.com/@rspesc31/forgot-to-add-node-modules-to-gitignore-solved-6c5b7a444c65)to find out how. Currently, I'm unable to view your code because your node_modules have consumed a significant amount of storage in your pull request.

client/src/AddVideo.jsx
```diff
@@ -0,0 +1,75 @@
...
+        console.log("Server response:", response.data); // Debugging line
+  
+        // If the video is added successfully on the server, update the client-side state
+        if (response.status === 201) {
+          onAdd(response.data); // Assuming response.data contains the new video data
+          setTitle('');
+          setUrl('');
+  
+          // Reload the page
+          window.location.reload();
```
illicitonion, line 33: You shouldn't need to reload pages in react apps - if you set new state, react will re-render the page for you if it needs it.

server/config/db.js
```diff
@@ -0,0 +1,23 @@
+const {Pool} = require("pg");
+const fs = require("fs");
+
+
```
illicitonion, line 4: You shouldn't add your password to source control - I'd recommend removing this, using a .env file, and changing your password.

server/server.js
```diff
@@ -1,15 +1,93 @@
...
+    console.error("Error adding video to the database", error);
+    res.status(500).json({ error: "Internal server error" });
+  }
+});
+
+
+
+app.get("/:id", (req, res) => {
+  const id = parseInt(req.params.id);
+  const video = videos.find((video) => video.id === id);
```
illicitonion, line 69: It looks like this didn't get updated to use the database? Same as the delete, below?

client/public/index.html
```diff
@@ -10,42 +10,58 @@
...
-      This HTML file is a template.
-      If you open it directly in the browser, you will see an empty page.
-
-      You can add webfonts, meta tags, or analytics to this file.
-      The build step will place the bundled scripts into the <body> tag.
 
-      To begin the development, run `npm start` or `yarn start`.
-      To create a production bundle, use `npm run build` or `yarn build`.
-    -->
+    <script>
```
illicitonion, line 30: What's the deal with this script here?

illicitonion (commented):
I removed your node_modules so I could review this code - it's generally looking good, but it looks like you have a few pieces of leftover code you maybe don't need any more?

PR #384, opened 2023-08-28
London10-AfshaHossain-FullStackProjectAssessment-Week2-Level 200

Afsha10:


client/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules/
```
giorgiguts, line 1: Great to see that node_modules is inside the .gitignore file. It's important that we abstract node_modules from Version Control.

client/src/App.js
```diff
@@ -1,13 +1,18 @@
...
+import youtubeLogo from "./components/images/youtubeLogo.png";
 
-function App() {
+const App = () => {
   return (
-    <div className="App">
-      <header className="App-header">
-        <h1>Video Recommendation</h1>
-      </header>
+    <div className="App text-left">
```
giorgiguts, line 8: If you are not using the `.App` as a CSS class, feel free to remove it. It's often best to de-clutter our codebase as much as possible.

client/src/components/Button.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+function Button({ onDelete, id }) {
```
giorgiguts, line 3: Great use of destructuring props here. 👍

client/src/components/CardsContainer.jsx
```diff
@@ -0,0 +1,43 @@
...
+1. pass the full data down to delete a small part from the full data
+2. instead of passing the full data down, we get the unique number of the card to delete it from the full data
+
+We turn the videoData into a useState to delete the video with a specific data.
+How do you delete a record from a useState array (answer might be spread operator, perhaps .map?)
+*/
+
+function CardsContainer( { videoData, setVideoData }) {
+
+function handleDelete(id) {
```
giorgiguts, line 17: what you can do as an additional measure of security is check if the `id` exists inside this function. 
Something along these lines:
  ```
if (!id) {
    // Handle the case where id is not provided or is invalid
    // You can return, throw an error, or take any appropriate action here.
    return;
  }
  
  setVideoData(
    videoData.filter((video) => {
      return video.id !== id;
    })
  );
```
  
The code is great as it is, but you can consider it for the future use.

File: client/src/components/MainContainer.jsx [hunk 5]
```diff
@@ -0,0 +1,18 @@
+import React, { useState } from "react";
+import CardsContainer from "./CardsContainer";
+import VideoForm from "./VideoForm";
+import fullData from "../exampleresponse.json";
+
+function MainContainer() {
+    const [videoData, setVideoData] = useState(fullData);
+    
+    console.log("Video Data from MainContainer---> ", videoData);
```
giorgiguts, line 9: When you push to the origin (on the cloud), it's worth looking over your codebase to check if there are any redundant development-level code lines scattered somewhere. For example, in most cases, console.log() helps us debug our program during the development phase (when we are actually coding and building the program/app). It's best if you remove any console.logs from your program to declutter the codebase.

client/src/components/VideoCard.jsx
```diff
@@ -0,0 +1,66 @@
+import React from "react";
+import Button from "./Button";
+
+const VideoCard = ({ videoData, setVideoData, singleVideo, onDelete }) => {
+  /**
+    On the return statement of the map, for the when the condition is true, I need to create the appropriate value that will replace the original value. When the condition is false I can return the video.
+   */
+
+  function upVoteHandler() {
+    console.log("videoData--->", videoData);
```
giorgiguts, line 10: Same here. Better to remove it before we push it to the cloud.

client/src/components/VideoForm.jsx
```diff
@@ -0,0 +1,92 @@
...
+    function validateUrl(urlObject) {
+      console.log("urlObject --->", urlObject);
+      if (!urlObject.startsWith("https://www.youtube.com")) {
+        return false;
+      }
+      return true;
+      // return either validated URL or an error
+    }
+
+    event.preventDefault();
```
giorgiguts, line 19: Usually, **preventDefault()** is better to move to the top, just after the function addVideoHandler starts. This way, you minimise the potential bugs from rising. 

```
 function addVideoHandler(event) {
     event.preventDefault();
     // code here...
 }
```
It's worth having a read about what it does and why we need to invoke it. Here's the MDN [link](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)

server/server.js
```diff
@@ -1,15 +1,71 @@
+require("dotenv").config();
 const express = require("express");
 const app = express();
 const port = process.env.PORT || 5000;
+const cors = require("cors");
```
giorgiguts, line 5: Not sure which version of Node you are using, but going forward I think it's better if you get acquainted with the `import/export` modules that was introduced in ES6. Here's a comprehensive [guide](https://www.digitalocean.com/community/tutorials/js-modules-es6) on how to start using them and why are they so powerful.

giorgiguts (commented):
Great work! 😊 
Keep in mind to remove those console.logs() from the codebase before pushing, as they no longer serve the purpose, once you've effectively debugged your program. Otherwise, amazing stuff and keep up the good work!

server/server.js
Afsha10: Thanks Giorgi for sharing about ES6! My Node version is 18.17.1. I will read it.

client/src/components/VideoForm.jsx
Afsha10: This is very useful to know. Thank you!

client/src/components/MainContainer.jsx
Afsha10: I decluttered them. Thank you for your feedback!

client/src/components/CardsContainer.jsx
Afsha10: Thank you!

client/src/App.js
Afsha10: I have removed the import related to CSS and also removed App.css file as I am not using CSS.

Afsha10:
> Great work! 😊 Keep in mind to remove those console.logs() from the codebase before pushing, as they no longer serve the purpose, once you've effectively debugged your program. Otherwise, amazing stuff and keep up the good work!

Thank you very much for taking the time to review my code! And your valuable feedback!

PR #390, opened 2023-08-30
London10-Bedrije_Omuri-Level-150

Bedi06:


client/src/components/AddVideo.js
```diff
@@ -0,0 +1,89 @@
+import React, { useState } from "react";
+
+function AddVideo({ onAddVideo }) {
+  const [title, setTitle] = useState("");
+  const [url, setUrl] = useState("");
+  const [Error, setError] = useState("");
+
```
ShayanMahnam, line 7: you can also set 0 vote here for new videos, now when user add new video they cant vote for it becuase its undefined.

client/src/components/AddVideo.js
```diff
@@ -0,0 +1,89 @@
...
+    event.preventDefault();
+
+    if (!title.trim()) {
+      setError("Please enter a title"); //title error handling
+    } else if (!validUrl(url)) {
+      setError("Please enter a valid YouTube URL"); //url error handling
+    } else {
+      onAddVideo({ title, url });
+      setTitle("");
+      setUrl("");
```
ShayanMahnam, line 31: you can use setError("") here as well so we will clear the error if the form submited correctly

ShayanMahnam (commented):
Well done so far Bedi, very neat code and easy to review. keep doing your amazing job!

PR #396, opened 2023-09-04
London_Class10_Bedrije-Omuri_Level-299

Bedi06:


client/src/components/VideoCards.js
```diff
@@ -0,0 +1,80 @@
+import React, { useState } from "react";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+import { faThumbsUp, faThumbsDown } from "@fortawesome/free-solid-svg-icons";
+
+function VideoCard({ title, url, rating, onRemove }) {
+  const videoId = url.split("v=")[1];
```
illicitonion, line 6: Is it possible a URL didn't contain `v=`? What would happen here if this happened? Is there a way that could be handled more smoothly?

server/server.js
```diff
@@ -1,15 +1,157 @@
...
+  },
+];
 
 // GET "/"
 app.get("/", (req, res) => {
-  // Delete this line after you've confirmed your server is running
-  res.send({ express: "Your Backend Service is Running" });
+  const { order } = req.query;
+  let orderedVideos = [...videos];
+  if (order === "asc") {
```
illicitonion, line 81: It looks like you have sorting both on the server side and the client side - can you talk through why you added it in both places? Are they doing different things, or the same thing? Could one be removed?

server/server.js
```diff
@@ -1,15 +1,157 @@
...
+});
+
+// GET "/{id}"
+app.get("/:id", (req, res) => {
+  const id = parseInt(req.params.id);
+
+  const video = videos.find((v) => v.id === id);
+
+  if (!video) {
+    res.status(404).json({
```
illicitonion, line 97: This structured error, where you return an object which contains both overall status (result) and description where needed is really nice - it helps to be able to give more information!

I'd suggest you maybe also want to use this structure for successful results:
```json
{
  "result": "success",
  "video": {
    "id": ...,
    "title": ...,
    "url": ...,
    "rating": ...
  }
}
```

That way you can always safely look at the "result" field and know whether it was successful or not. Imagine if your `video` object in the future gets a property called "result", that could be confusing.

server/server.js
```diff
@@ -1,15 +1,157 @@
...
+  const { title, url } = req.body;
+
+  if (
+    !title ||
+    !url ||
+    !url.match(/^(https?:\/\/)?(www\.)?youtube\.com\/watch\?v=[\w-]+(&\S+)?$/)
+  ) {
+    res.send({
+      result: "failure",
+      message: "Video could not be saved",
```
illicitonion, line 118: I'd recommend including a reason about _why_ here, which you can show to the user. You know what the problem is - either the title is empty, the URL is empty, or the URL is bad - tell the user what specifically they did wrong so they know how to fix it :)

server/server.js
```diff
@@ -1,15 +1,157 @@
...
+    });
+  }
+
+  let newId = 0;
+  while (videos.some((video) => video.id === newId)) {
```
illicitonion, line 123: There are couple of really interesting things about this way of choosing an ID:
1. You may re-use them - if you add a video, then delete it, a new video may get its ID - can you think of any places this may be problematic?
2. (Advanced topic, and I'm very happy to chat about this if you want) It involves lots of searching through the videos list, and it will slow down a lot as you get more videos. Let's imagine you had 1000 videos with IDs 0..999 - you're going to look through all 1000 videos, 1000 times, so you're going to do 1 million "does this video already have this ID" checks. Can you think of a way that we could do this faster?

illicitonion (commented):
This is looking really good, congrats!

PR #397, opened 2023-09-05
London 10 - Elena Barker - Full - Stack - Project - Assessment - Level 200

ElenaBarker:


sherif98 (commented):
Can you change the PR to make the base branch Level-100? Currently, the PR is showing diff from main to Level-200. Ideally, it should display only diff between Level-100 and Level-200 branches.

Other than that I think the server changes look good.

PR #399, opened 2023-09-05
London10-AfshaHossain-FullStackProjectAssessment-Week2-Level 250

Afsha10:


client/src/components/Button.jsx
```diff
@@ -0,0 +1,14 @@
+import React from "react";
+
+function Button({ onDelete, id }) {
```
illicitonion, line 3: A couple of things to think about with this component:
1. "Button" is a very general name - can you think of a more specific name for what the button actually is for?
2. `onDelete` is a slightly strange name - "on" as a prefix normally means this function is meant to be called when something has happened (e.g. onClick is for when something has been clicked), but this actually _does_ the delete, rather than handles it. Can you think of a name which suggests more this _does_ the delete, rather than is called as a _result_ of the delete?

client/src/components/CardsContainer.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+function CardsContainer({ videoData, setVideoData, setFetchData }) {
+  async function handleDelete(id) {
+    fetch(`http://localhost:5000/videos/${id}`, {
+      method: "delete",
+    })
+      .then((response) => response.json())
+      .then((result) => console.log(result))
+      .catch((error) => console.log("error", error));
+    setFetchData(true);
```
illicitonion, line 12: What is `setFetchData` for? It's not super clear from its name what it's for, or what "true" vs "false" means for its value?

client/src/components/MainContainer.jsx
```diff
@@ -0,0 +1,34 @@
+import React, { useEffect, useState } from "react";
+import CardsContainer from "./CardsContainer";
+import VideoForm from "./VideoForm";
+
+function MainContainer() {
+  const [videoData, setVideoData] = useState();
+  const [fetchData, setFetchData] = useState(true);
```
illicitonion, line 7: When is this state ever `false`? What does this state control?

client/src/components/VideoCard.jsx
```diff
@@ -0,0 +1,77 @@
...
+          },
+          body: JSON.stringify(video),
+        });
+      }
+      return video;
+    });
+    setVideoData(likeRating);
+  }
+
+  function downVoteHandler() {
```
illicitonion, line 26: This function has a lot in common with the `upVoteHandler` - can you think how to make them share code, rather than having two copies of basically the same code?

client/src/components/VideoForm.jsx
```diff
@@ -0,0 +1,100 @@
...
+          urlObject.startsWith("https://youtu.be") ||
+          urlObject.startsWith("https://m.youtube.com") ||
+          urlObject.startsWith("https://youtube.com/"))
+      );
+    }
+
+    const formTitle = event.target.form.title.value;
+    const formUrl = event.target.form.url.value;
+
+    const cleanedUrl = formUrl.includes("&") ? formUrl.split("&")[0] : formUrl;
```
illicitonion, line 22: What's this code cleaning up? What does this code achieve? Maybe worth a comment, with an example of what's being removed, just above the code :)

client/src/components/VideoForm.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+    if (validateUrl(formUrl) && formTitle !== "") {
+      const newData = {
+        id: parseInt(Math.random() * 1000),
```
illicitonion, line 26: 1000 is a pretty small number of IDs, and is likely to have collisions (multiple videos with the same ID) with only a few videos. (See https://en.wikipedia.org/wiki/Birthday_problem for some interesting information on the probabilities here!)

Also, the server should probably be picking the IDs, not the client, as the server is the thing that knows what IDs are already used.

client/src/components/VideoForm.jsx
```diff
@@ -0,0 +1,100 @@
...
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(newData),
+      });
+
+      setFetchData(true);
+      setErrorMessage("");
+    } else if (formTitle === "" && !validateUrl(formUrl)) {
+      setErrorMessage("Add a title and a valid URL");
```
illicitonion, line 45: Good job making sure the page is clear about what the specific problem was with the form!

client/src/components/VideoForm.jsx
```diff
@@ -0,0 +1,100 @@
...
+    } else if (formTitle === "") {
+      setErrorMessage("Add a title");
+    } else {
+      setErrorMessage("Add a valid URL");
```
illicitonion, line 49: It may be worth describing what a valid URL looks like (e.g. if I put in https://github.com, that's valid URL - you have stricter criteria than that, but you're not telling me what they are :))

server/server.js
```diff
@@ -1,15 +1,83 @@
...
-app.get("/", (req, res) => {
-  // Delete this line after you've confirmed your server is running
-  res.send({ express: "Your Backend Service is Running" });
+app.get("/videos", (req, res) => {
+  res.status(200).json(videos);
+});
+
+app.post("/videos", (req, res) => {
+  const { title, url } = req.body;
+  if (!title || !url || !url.startsWith("https://www.youtube.com")) {
```
illicitonion, line 19: Your client had different validation than your server, for what URLs are allowed - this may be a problem!

server/server.js
```diff
@@ -1,15 +1,83 @@
...
+    res.status(400).json({
+      result: "failure",
+      message: "Video could not be saved",
```
illicitonion, line 22: It'd be handy to include in the error message what was wrong with the request, so that a more specific error can be presented to the user.

server/server.js
```diff
@@ -1,15 +1,83 @@
...
+app.delete("/videos/:id", (req, res) => {
+  let id = Number(req.params.id);
+  const matchingVideo = videos.find((video) => {
+    return video.id === id;
+  });
+
+  if (!matchingVideo) {
+    res.status(400).json({
+      result: "failure",
+      message: "Video could not be deleted",
```
illicitonion, line 59: It may be worth explaining why (that the video didn't exist/wasn't know), otherwise the user may think the website is broken, retry, and be confused.

server/server.js
```diff
@@ -1,15 +1,83 @@
...
+    const videoIndexToBeDeleted = videos.indexOf(matchingVideo);
+    videos.splice(videoIndexToBeDeleted, 1);
+    res.status(200).json({});
+  }
+});
+
+app.put("/videos/:id", (req, res) => {
+  const newVideo = req.body;
+  let id = Number(req.params.id);
+  const videoIndex = videos.findIndex((video) => {
```
illicitonion, line 72: In the previous function you used `videos.indexOf`, and here you're using `videos.findIndex` - is there a difference? Why did you choose one and the other?

illicitonion (commented):
This is looking good, congrats!

client/src/components/Button.jsx
Afsha10: Thank you @illicitonion for pointing these out and paying attention to details! Having the right names is super helpful for me to understand my own code when I look at them.
After going through your review, I made changes to my level 300 as my level 300 is built on top of my level 250. 
1. I have renamed my Button component to DeleteButton.
2. onDelete does sound strange. I have renamed it to handleDelete.

client/src/components/CardsContainer.jsx
Afsha10: I have removed setFetchData and added a new useState variable called "setVideoAdded" and set it as false and true when adding new videos data.

client/src/components/VideoCard.jsx
Afsha10: I have a new function in my level 300 called vote which I used to connect with both the upVote and downVote handlers.

client/src/components/VideoForm.jsx
Afsha10: Thank you for reminding me to put that comment! I have added the comment with an example. I had to remove the end bits so that I get to display the thumbnails of the YouTube videos in my app.

client/src/components/VideoForm.jsx
Afsha10: I have changed my error message to, "Add a valid YouTube video URL". :))

server/server.js
Afsha10: I have added  a URL validation in my level 300 server.js today.

server/server.js
Afsha10: Corrected my error messages in level 300. :)

server/server.js
Afsha10: Corrected this message in level 300.

PR #401, opened 2023-09-07
Glasgow 6 | Artem Filkovskyi | Full Stack Assessment

filkovskyy:


Dedekind561 (commented):
Hi Artem, 
Thanks for your submission! I'm seeing lots of good work here.
I think something strange is going on with the CSS at the moment though: you seem to be passing inline CSS using the `className` property. Is this just a mistake?

PR #402, opened 2023-09-07
LONDON-10/BAKI TUNCER/ FULL-STACK-PROJECT

batuncer:
NAME:BAKI TUNCER
LONDON-10
FULLSTACK PROJECT


ShadiSafaee:
Hi @batuncer  - could you add node_modules to .gitignore file? You can follow [this ](https://medium.com/@rspesc31/forgot-to-add-node-modules-to-gitignore-solved-6c5b7a444c65)to find out how to do so. Currently, I'm unable to view your code because your node_modules have consumed a significant amount of storage in your pull request.

client/src/components/AddVideo.jsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  const handleAddVideo = () => {
+    if (title.trim() === '' || !isValidYouTubeUrl(url)) {
+      alert('Please enter a valid title and a valid YouTube URL.');
+      return;
+    }
+
+    // Validate if the URL is a YouTube video URL and extract the video ID
+    const videoId = getYouTubeVideoIdFromUrl(url);
```
illicitonion, line 33: What happens if this returns `null`?

client/src/components/AddVideo.jsx
```diff
@@ -0,0 +1,75 @@
...
+    // Call the parent component's function to add the video
+    onAddVideo(newVideo);
+
+    // Reset the input fields
+    setTitle('');
+    setUrl('');
+  };
+
+  // Function to check if a string is a valid YouTube URL
+  const isValidYouTubeUrl = (url) => {
```
illicitonion, line 53: It feels like your `getYouTubeVideoIdFromUrl` function is doing better validation than this function - is this function really worth having/calling?

server/server.js
```diff
@@ -1,15 +1,98 @@
...
-let videos = [];
+let videos = exampledata;
 
 // GET "/"
 app.get("/", (req, res) => {
-  // Delete this line after you've confirmed your server is running
-  res.send({ express: "Your Backend Service is Running" });
+  const order = req.query.order || "desc";
+
+  // Sort videos based on the 'order' parameter
```
illicitonion, line 28: It looks like you're sorting both on the server side and the client side - do you need to be doing both? What advantages/disadvantages does each place to do it have?

server/server.js
```diff
@@ -1,15 +1,98 @@
...
+app.post("/", (req, res) => {
+  const { title, url } = req.body;
+
+  // Validate title and URL
+  if (!title || !url) {
+    return res.status(400).json({ result: "failure", message: "Title and URL are required." });
+  }
+
+  // Generate a unique ID for the new video
+  const id = nextVideoId++;
```
illicitonion, line 48: Where is nextVideoId defined? Can you talk through how this code works / what it does? What's the scope of `nextVideoId`?

illicitonion (commented):
This is generally looking pretty good, but there are a few pieces of detail that seem slightly off (e.g. calling your extraction function which may return `null`, but then not handling `null`). Also, did you manage to do the database parts (level 300)?

PR #404, opened 2023-09-08
glasgow_6|Najah_Farah|fullstack_assessment-level 300

nfarah22:


Dedekind561:
@nfarah22 You've currently committed your `node_modules` - please remove these so we can review your code. 
In addition, please post a link to your deployed application too!
Thanks

PR #417, opened 2023-09-13
Glasgow class 6- Siver Omar- Full Stack Project Assessment level 300

siveromar:


client/src/VideoContainer.js
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from "react";
+
+class VideoContainer extends Component {
```
AightA, line 3: is this require to be class component ? have you tried functional method ?

AightA (commented):
LGTM - looks good to me.

PR #418, opened 2023-09-13
London10-Anna_Hrychaniuk-Level300

AnnaFYZ:


client/src/App.js
```diff
@@ -1,11 +1,135 @@
...
+        console.log("Response:", data);
+        setUpdated(!updated);
+      })
+      .catch((error) => {
+        console.error("Error:", error);
+      });
+  };
+
+  const upVote = (video) => {
+    const updatedVideos = videos.map((v) =>
```
kitko112, line 42: Nice logic that only change the vote of a particular video.

FYI do you know the difference between `by reference` and `by value`?
https://stackoverflow.com/a/6605700

client/src/App.js
```diff
@@ -1,11 +1,135 @@
...
+          onClick={() => setSortOrder(sortOrder === "desc" ? "asc" : "desc")}
+        >
+          Toggle Sort Order
+        </button>
+      </div>
+
+      <ErrorBoundary
+        fallback={<div>Could not connect to the server, sorry....</div>}
+      >
+        {loading && "Loading..."}
```
kitko112, line 111: To write a more readable code, we can move this part above the return and write it as below

```
if(isloading){
  return <>Loading...</>
}


```

then you dont need !loading checking on line 112

client/src/Videocard.js
```diff
@@ -0,0 +1,55 @@
+import { library } from "@fortawesome/fontawesome-svg-core";
+import { faThumbsUp, faThumbsDown } from "@fortawesome/free-solid-svg-icons";
+import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
+
+library.add(faThumbsUp, faThumbsDown);
+
+function Videocard({ video, onUpVote, onDownVote, onRemove }) {
+  let linkFixed = video.url.replace("watch?v=", "embed/");
```
kitko112, line 8: if this line can be done after the fetch, this Videocard component can purely focus on the presentation only.

Image one day you want to replace the Videocard component but you forget to add this line, it would be nice to keep the data logic with set useState, have cleaner separation of logic

server/server.js
```diff
@@ -1,15 +1,129 @@
...
+//       "The Complete London 2012 Opening Ceremony | London 2012 Olympic Games",
+//     url: "https://www.youtube.com/watch?v=4As0e4de-rI",
+//     rating: 3211,
+//   }
+// ];
+
+// GET "/:{id}"
+app.get("/:id", (req, res) => {
+  let idRequested = Number(req.params.id);
+  db.query('select * from urls where id = $1', [idRequested])
```
kitko112, line 56: if the idRequested is undefined, we can return an HTTP 404 early and skip the call to database, does it sound better?

server/server.js
```diff
@@ -1,15 +1,129 @@
...
+  .catch(err => res.send(err));
+  // let videoRequested = videos.find(video => video.id === idRequested);
+  // videoRequested
+  // ? res.status(200).json(videoRequested)
+  // : res.status(400).send("Not found")
+});
+
+app.post("/update_rating", (req, res) => {
+  const {id, rating} = req.body;
+  db.query('update urls set rating = $1 where id = $2', [rating, id])
```
kitko112, line 67: Same as above comment, we can validate the params first before we pass it to database

server/server.js
```diff
@@ -1,15 +1,129 @@
...
 
-// Store and retrieve your videos from here
-// If you want, you can copy "exampleresponse.json" into here to have some data to work with
-let videos = [];
 
 // GET "/"
 app.get("/", (req, res) => {
-  // Delete this line after you've confirmed your server is running
-  res.send({ express: "Your Backend Service is Running" });
+  db.query('select * from urls')
```
kitko112, line 82: It should work in this assignment.

here is a real life scenario, if we have millions of video, who do we make calls to retrieve the videos? you can think about it after you have graduated

server/server.js
```diff
@@ -1,15 +1,129 @@
...
+  .then(result => res.json(result.rows))
+  .catch(err => res.send(err));
 });
+
+//POST "/"
+app.post("/", (req, res) => {
+  const { title, url, dateTime } = req.body;
+  db.query(
+    "insert into urls (title, url, rating, timeAdded) values ($1, $2, 0, $3) returning id", [title, url, dateTime]
```
kitko112, line 91: Appreciate you pass the params to db query via $1, $2, it is the right way to avoid the sql injection https://owasp.org/www-community/attacks/SQL_Injection

.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+node_modules
```
illicitonion, line 2: I notice you have `node_modules` in here twice - can you talk through what having it twice does different from once?

client/src/App.js
```diff
@@ -1,11 +1,135 @@
...
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ id: video.id, rating: video.rating + 1 }),
+    })
+      .then((response) => response.json())
+      .then((data) => console.log(data))
+      .catch((error) => console.log("Error: ", error));
+  }
+
+  const downVote = (video) => {
```
illicitonion, line 59: This function looks very similar to the `upVote` function - can you think of a way to make them share code rather than duplicate it?

server/server.js
```diff
@@ -1,15 +1,129 @@
...
+//     url: "https://www.youtube.com/watch?v=4As0e4de-rI",
+//     rating: 3211,
+//   }
+// ];
+
+// GET "/:{id}"
+app.get("/:id", (req, res) => {
+  let idRequested = Number(req.params.id);
+  db.query('select * from urls where id = $1', [idRequested])
+  .then(result = res.status(200).json(result.rows))
```
illicitonion, line 57: In general I would expect an endpoint to get `/id` to just return a single object, rather than array containing a single object, and to 404 if the ID was missing - what do you think?

illicitonion (commented):
Looks great, congrats!

PR #419, opened 2023-09-14
Glasgow Class 6 - Rahma Berhan - Full Stack Assessment - Level 300 

RahmaB1:


Dedekind561:
@RahmaB1 do you have a link to the deployed version of your app?

PR #428, opened 2023-09-15
London 10 | Elena Barker | Full-Stack-Project | Level-300

ElenaBarker:


client/src/CountVotes.js
```diff
@@ -0,0 +1,71 @@
+import { useState } from "react";
+import ThumbUp from "@material-ui/icons/ThumbUp";
+import ThumbDown from "@material-ui/icons/ThumbDown";
+
+const CountVotes = () => {
+  const [likeCount, setLikeCount] = useState(0);
```
illicitonion, line 6: These three pieces of state all only ever change together, which means they maybe could be collapsed into one piece of state (maybe with values "like", "neutral", and "dislike") - what do you think?

server/.env
```diff
@@ -0,0 +1 @@
+DB_URL = postgres://elena:EryWSIiXXDFufKrilWGtm1BOeO2tOrFH@dpg-ck27hvn03lhc7389didg-a.frankfurt-postgres.render.com/dbvideoselena
```
illicitonion, line 1: You probably didn't mean to check this file in - I'd remove it, add it to your `.gitignore`, and change your password :)

server/server.js
```diff
@@ -1,15 +1,158 @@
...
+dotenv.config();
+const { Pool } = require("pg");
+
+const db = new Pool({
+  connectionString: process.env.DB_URL,
+  ssl: { rejectUnauthorized: false },
+});
+const bodyParser = require("body-parser");
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(bodyParser.json());
```
illicitonion, line 16: I think the second call here cancels out the first one - there's only one kind of body parser you're registering, and it will use the last one you registered - you probably want to remove the first line here.

server/server.js
```diff
@@ -1,15 +1,158 @@
...
+    .then(() => {
+      res.status(201).send("New video succesfully added!");
+    })
+    .catch((error) => {
+      console.log(error);
+    });
+});
+
+// GET "/{id}"
+app.get("/:id", (req, res) => {
```
illicitonion, line 136: It looks like this endpoint and the `/videos/:id` endpoint didn't get updated to use the database yet?

illicitonion (commented):
This is generally looking great, and the code is really clear and easy to read - good job!

server/server.js
```diff
@@ -1,15 +1,158 @@
...
+  }
+});
+
+// DELETE "/{id}"
+
+app.delete("/videos/:id", (req, res) => {
+  const videoId = parseInt(request.params.id);
+  const singleVideo = videos.find((singleVideo) => singleVideo.id === videoId);
+  if (singleVideo) {
+    videos.splice(singleVideo, 1);
```
illicitonion, line 155: Splice takes as its first argument an _index_ not a value - are you sure this code works?

server/.env
ElenaBarker: Done. Thank you)

server/server.js
ElenaBarker: Thank you. I deleted the first line)

PR #435, opened 2023-09-18
London-10 | Iryna Lypnyk | Full Stack Assessment | Level 300

IrynaLypnyk:


client/src/App.js
```diff
@@ -1,13 +1,97 @@
...
+    useEffect(() => {
+        getAllVideos(order, setVideos, setLoading);
+    }, [order])
+
+
+    const handleDeleteVideo = (id) => {
+        try {
+            fetch(`${API_URL}/${id}`, {
+                method: "DELETE", // или 'PUT'
+                body: JSON.stringify({id}), // данные могут быть 'строкой' или {объектом}!
```
illicitonion, line 35: You generally don't need to include the same data in both the path and the body - you can probably just put it in one place or the other, as long as the server knows where it's going to read it from :)

client/src/App.js
```diff
@@ -1,13 +1,97 @@
...
+                    .then(() => {
+                        getAllVideos(order, setVideos, setLoading);
+                    });
+            } catch (error) {
+                console.error({error});
+            }
+        }
+    };
+
+    const handleOrderChange = (value) => {
```
illicitonion, line 67: I'm not sure what value this function is adding over just calling `setOrder` directly?

client/src/components/Video/index.js
```diff
@@ -0,0 +1,65 @@
...
+import React, {useEffect} from "react";
+import {API_URL} from "../../App";
+import styles from "./index.module.css";
+
+const {Text} = Typography;
+const Video = (props) => {
+    const {video: {id, title, url, rating: videoRating, timestamp}, handleDeleteVideo} = props;
+    const [rating, setRating] = React.useState(videoRating);
+
+    const videoId = url.split("v=")[1].substring(0, 11);
```
illicitonion, line 11: What's special about the first 11 characters? It may be worth including a comment describing why you're stripping off the end, and why that's the right place to stop.

illicitonion (commented):
This is looking great, good job!

PR #436, opened 2023-09-18
London class-10 | Junita lama | full-stack-assessment | level 250

Junitalama:


client/src/components/AllButtons.jsx
```diff
@@ -0,0 +1,76 @@
...
+  const [videos, setVideos] = useState([]);
+
+  useEffect(() => {
+    fetch(Api)
+      .then((response) => response.json())
+      .then((data) => {
+        console.log("Fetched Data:", data);
+        const videosWithRating = data.map((video) => ({
+          ...video,
+          rating: video.rating || 0,
```
illicitonion, line 18: Why do you need to do this? What situation leads to you getting a video with no rating?

server/server.js
```diff
@@ -1,15 +1,76 @@
...
 
 // GET "/"
 app.get("/", (req, res) => {
-  // Delete this line after you've confirmed your server is running
-  res.send({ express: "Your Backend Service is Running" });
+  res.json(videos);
+});
+
+app.post("/", (req, res) => {
+  const newId = videos.length + 1;
```
illicitonion, line 20: Are you sure this will always be a unique ID? Can you think of a time when this technique would end up using the same ID for two different videos?

server/server.js
```diff
@@ -1,15 +1,76 @@
...
+    url: req.body.url,
+    rating:0,
+  };
+  if (req.body.title && req.body.url) {
+    videos.push(video);
+    res.status(200).send(video);
+  } else
+    res.status(400).json({
+      result: "failure",
+      message: "Video could not be saved",
```
illicitonion, line 33: I'd include in this message _why_, so you can present that error to the user and tell them how to fix their problem.

server/server.js
```diff
@@ -1,15 +1,76 @@
...
+    });
+});
+
+app.get("/:id", (req, res) => {
+  const id = Number(req.params.id);
+  const filteredVideos = videos.filter((video) => video.id === id);
+  if (filteredVideos.length > 0) {
+    res.status(200).json(filteredVideos);
```
illicitonion, line 41: In general when I ask an endpoint for a single video, I'd expect it to return one object, rather than an array containing one object - can you return just an object here?

server/server.js
```diff
@@ -1,15 +1,76 @@
...
+app.delete("/:id", function (req, res) {
+  let id = Number(req.params.id);
+  let deleteIndex = videos.findIndex((video) => video.id === id);
+  if (deleteIndex >= 0) {
+    videos.splice(deleteIndex, 1);
+    res.status(200).json({});
+  } else {
+    res.status(404).json({
+      result: "failure",
+      message: "Video could not be deleted",
```
illicitonion, line 59: I think "video could not be found" is probably a more clear error message than "video could not be deleted" :)

server/server.js
```diff
@@ -1,15 +1,76 @@
...
+    });
+  }
+});
+
+app.put("/:id", function (req, res) {
+  let id = Number(req.params.id);
+  let videoIndex = videos.findIndex((video) => video.id === id);
+  if (videoIndex >= 0) {
+    videos[videoIndex].rating = Number(req.body.rating);
```
illicitonion, line 68: Do you want to allow the user to specify arbitrary numbers in one request, or would you rather only allow them to add/remove one to the rating per request? What advantages/disadvantages does each approach have?

illicitonion (commented):
This is generally looking good, nicely done! It looks like your level 300 PR is missing all of its content, though? https://github.com/CodeYourFuture/Full-Stack-Project-Assessment/pull/437

PR #447, opened 2023-09-23
London_10 | Yulia - Hospodar | Full-Stack-Project |  

Yuliya110692:


client/src/components/SingleVideo.js
```diff
@@ -0,0 +1,63 @@
+import { useState } from "react";
+import { FaThumbsUp, FaThumbsDown } from "react-icons/fa";
+import "./SingleVideo.css";
+
+const SingleVideo = ({ video, handleDeleteVideoClick }) => {
+  const { title, url, rating, id } = video;
+  const [votes, setVotes] = useState(0); // can be 0, 1 or -1 => to add to rating when thumbsUp or ThumbsDown is clicked
```
illicitonion, line 7: It's really good to include a comment explaining what values are allowed/possible when ony some values are allowed - good job!

client/src/components/SingleVideo.js
```diff
@@ -0,0 +1,63 @@
...
+
+  return (
+    <div className="video-card">
+      <h3>{title}</h3>
+      <div className="iframe-container">
+        <iframe
+          className="responsive-iframe"
+          width="560"
+          height="310"
+          src={`https://www.youtube.com/embed/${url.split("v=")[1]}`}
```
illicitonion, line 27: What would happen if a URL _didn't_ contain `v=`? Is there a way you could handle that better?

client/src/components/SingleVideo.js
illicitonion: That said, I don't see any code that _prevents_ votes from being set to 2, or -2 - do you intend for there to be code stopping that?

client/src/components/Videos.js
```diff
@@ -0,0 +1,98 @@
...
+    var p =
+      /^(?:https?:\/\/)?(?:m\.|www\.)?(?:youtu\.be\/|youtube\.com\/(?:embed\/|v\/|watch\?v=|watch\?.+&v=))((\w|-){11})(?:\S+)?$/;
+    if (url.match(p)) {
+      return url.match(p)[1];
+    }
+    return false;
+  };
+
+  const handleAddVideo = async (title, url) => {
+    if (title !== "" && url !== "" && matchYoutubeUrl(url)) {
```
illicitonion, line 41: This works, but often times we'll write this like:
```js
if (title === "" || url === "" || !matchYoutubeUrl(url)) {
  setShowError(true);
  return;
}
try {
  ...
}
```

Can you talk through why that may be more clear?

server/db.js
```diff
@@ -0,0 +1,15 @@
+const { Pool } = require('pg');
+
+link = "postgres://vtsohdxm:jNSKq4-96QSdfZax4qdrprDhVwu-rcrF@surus.db.elephantsql.com/vtsohdxm"
```
illicitonion, line 3: You probably don't want to check this in, containing a password - can you use the `dotenv` library and a `.env` file to set this up instead?

server/server.js
```diff
@@ -1,15 +1,114 @@
...
+
+    res.status(201).json(result.rows[0]); 
+  } catch (error) {
+    console.error('Error creating video:', error);
+    res.status(500).json({ error: 'Internal server error' });
+  }
+});
+
+
+app.delete("/:id", (req, res) => {
```
illicitonion, line 67: Can you update this to work on the database?

server/server.js
```diff
@@ -1,15 +1,114 @@
...
+
+  console.log(req.body)
+  // validate the req
+  const schema = Joi.object({
+    rating: Joi.number().required()
+  });
+
+  const result = schema.validate(req.body);
+  console.log(result)
+  const foundVideo = videos.find((booking) => {
```
illicitonion, line 100: `booking` seems like a strange name for this parameter :)

server/server.js
```diff
@@ -1,15 +1,114 @@
...
+    res.status(200).send(`Deleted video with id: ${videoID}`);
+  } else {
+    res.status(404).send({
+      result: "failure",
+      message: "Video could not be deleted",
+    });
+  }
 });
+
+app.put("/:id", (req, res) => {
```
illicitonion, line 86: Can you update this to work on the database?

illicitonion (commented):
This is generally looking really good! I've left a few comments of things to think about / improve :)

PR #504, opened 2024-06-02
Improve test diagnostics

textbook:
The failing states of the tests weren't particularly useful, so:

- `expect(foo.length).toBe(bar)` -> `expect(foo).toHaveLength(bar)`, because the latter shows you what _was_ in `foo` if its length isn't `bar` - see e.g. [`jest/prefer-to-have-length`](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-have-length.md)
- `expect(foo[0].bar).toBe(baz)` -> `expect(foo).toEqual([{ bar: baz }])` again because then the failing test tells you what `foo` _was_, rather than _"expected `baz` received `undefined`"_ (or crashing with `TypeError: Cannot read properties of undefined`)
    - Another alternative, which may be useful in some circumstances, would be `expect(foo).toHaveProperty([0, "bar"], baz)`

[![][1]][1]

_From Growing Object-Oriented Software by Nat Pryce and Steve Freeman, © 2010 Nat Pryce, released under [CC BY-SA 4.0][2]._

Also:

- `\c videorec;` broke the test seeding
- Depending on the implementation, `await waitForElementToBeRemoved(deleteButton);` can fail because the same button is reused by React - wait for there to be only one button instead
- Use role-based selectors to give a11y feedback
- Use user-event interactions to let the tests focus on behaviour
- An idiomatic delete is 204 No Content, as the resource no longer exists to return

  [1]: http://www.growing-object-oriented-software.com/figures/feedback-on-diagnostics.svg
  [2]: https://creativecommons.org/licenses/by-sa/4.0/

### DIFF
--- a/client/src/App.test.jsx
+++ b/client/src/App.test.jsx
@@ -1,8 +1,4 @@
-import {
-	render,
-	screen,
-	waitForElementToBeRemoved,
-} from "@testing-library/react";
+import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { http, HttpResponse } from "msw";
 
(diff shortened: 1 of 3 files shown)


Dedekind561 (approved):
LGTM, thanks @textbook!

